PR #474, opened 2019-06-05
Update to font awesome 5

cdfa:
I started this to be able to use the `balance` icon, but I found `gavel` was a better alternative. Finished it anyway, because it's shame not to.

I suggest reviewing only the stylesheet file changes and going through all the pages to see if I missed any icons.


Riscky:
replaced by #595

PR #481, opened 2019-09-07
Updated the Gemfile

martijncasteel:
Fixed some vulnerabilties

image


maartenberg:
I think that all the changes in this PR are also in #482, can we close this?

martijncasteel:
@maartenberg Yes I think so too!

PR #490, opened 2019-10-15
Fix crash when searching for Members

### LINKED ISSUE #489: will_paginate broke user pagination if searching
On the current `development` branch, it's not possible to search for members, as causes a 500 due to `will_paginate` calling methods that don't exist.

The root cause of this is that the returned set of `Member`s is of a different type if `.search` is used, as Fuzzily returns an Array instead of a Rails result set, which will_paginate doesn't support. We apparently didn't test this.

maartenberg:
Fixes #489.

This PR will add a more basic pagination parameter for the search results, containing only 'next' and 'previous' buttons.

Also fixes an issue that caused the pagination numbers to render inside the column for the limit dropdown.

### DIFF
--- a/app/controllers/admin/members_controller.rb
+++ b/app/controllers/admin/members_controller.rb
@@ -7,9 +7,9 @@ class Admin::MembersController < ApplicationController
     @limit = params[:limit] ? params[:limit].to_i : 50
 
     if params[:search]
-      @members = Member.search(params[:search].clone)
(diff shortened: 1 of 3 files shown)


maartenberg:
>  However, on the payments page the table isn't immediately updated when searching.
Not sure if this was supposed to be fixed in here or if it's a separate issue.

We're fixing this in a separate issue

PR #493, opened 2019-11-01
International phone number support

### LINKED ISSUE #487: International phone numbers are not permitted
### Expected behaviour
An international student wants to sign up on the intro.svsticky.nl page and enters his phone number.
Everything goes well and the student is signed up.

### Actual behavior
The students enters his (international) phone number and gets an error on clicking submit.
An admin cannot set an non 10-digit phone number either

FInnvos123:
Fixes #487 

Note: 
The migration to convert phone numbers to E.164 format is irreversible as all invalid numbers are replaced with NULL.

### DIFF
--- a/app/assets/javascripts/intro.js
+++ b/app/assets/javascripts/intro.js
@@ -14,6 +14,8 @@
 //= require turbolinks/dist/turbolinks
 //= require bootstrap/dist/js/bootstrap
 //= require jquery.validate
+//= require intl-tel-input/build/js/intlTelInput-jquery
+//= require intl-tel-input/build/js/utils
 
 $(document).on('ready page:load turbolinks:load', function () {
   var disabledStudyOptions = null;
@@ -104,6 +106,20 @@ $(document).on('ready page:load turbolinks:load', function () {
     errorPlacement: function (error, element) {
     }
   });
+
+  $('#phone_input').intlTelInput({
+    preferredCountries: ['nl'],
+    separateDialCode: true,
+    hiddenInput: 'phone_number',
+    utilsScript: 'utils.js'
+  });
+
+  $('#emergency_phone_input').intlTelInput({
+    preferredCountries: ['nl'],
+    separateDialCode: true,
+    hiddenInput: 'emergency_phone_number',
+    utilsScript: 'utils.js'
+  })
 
   $('select#method').on("change", function () {
     if ($(this).val() == 'Cash/PIN') {
--- a/app/assets/javascripts/intro.js
+++ b/app/assets/javascripts/intro.js
@@ -14,7 +14,7 @@
 //= require turbolinks/dist/turbolinks
 //= require bootstrap/dist/js/bootstrap
 //= require jquery.validate
-//= require intl-tel-input/build/js/intlTelInput-jquery
+//= require intl-tel-input/build/js/intlTelInput
 //= require intl-tel-input/build/js/utils
 
 $(document).on('ready page:load turbolinks:load', function () {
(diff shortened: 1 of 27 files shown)


app/views/admin/members/index.html.haml
```diff
@@ -52,7 +52,7 @@
               %td= link_to "#{member.infix}", member
               %td= link_to "#{member.last_name}", member
               %td
-                = link_to "tel:#{member.phone_number}" do
+                = link_to "tel:#{TelephoneNumber.parse(member.phone_number).international_number}" do
```
maartenberg: If we enforce that every phone number in the database is in the international format, do we then also need to parse all the numbers again when showing them?

app/views/admin/members/show.html.haml
```diff
@@ -41,9 +41,9 @@
               %br
               #{@member.postal_code} #{@member.city}
             %p
-              #{@member.phone_number}
+              #{TelephoneNumber.parse(@member.phone_number).international_number()}
```
maartenberg: Style nitpick: When calling a function with no arguments, you can leave out the `()` at the end (like properties in Python).

app/assets/stylesheets/intl_tel_input.css
```diff
@@ -0,0 +1,4 @@
+/* Make international telephone input use the whole column */
```
maartenberg: Each of the usecases of this file requires `intl-tel-input/build/css/intlTelInput` to be used as well, so it might be better if we make this file `*= require` that file here.

app/assets/javascripts/intro.js
```diff
@@ -105,6 +107,45 @@ $(document).on('ready page:load turbolinks:load', function () {
...
+    hiddenInput: 'phone_number',
+    utilsScript: 'utils.js'
+  });
+
+  var iti_emergency_phone_input = window.intlTelInput(emergency_phone_input, {
+    preferredCountries: ['nl'],
+    separateDialCode: true,
+    hiddenInput: 'emergency_phone_number',
+    utilsScript: 'utils.js'
+  });
```
maartenberg: This bit seems to be almost the same as [this bit](https://github.com/svsticky/constipated-koala/pull/493/files#diff-e6404b253eda092a7b6fec30c7aa84a8R7-R23), except for the id of the inputs. Can we deduplicate this?

app/assets/javascripts/intro.js
```diff
@@ -105,6 +107,45 @@ $(document).on('ready page:load turbolinks:load', function () {
...
+
+  function validate_phone_number(input, instance) {
```
maartenberg: Can you add a comment on why this is necessary?

app/views/admin/members/index.html.haml
FInnvos123: I'm removing this because it doesn't change the formatting in any way.

It's not required to parse the numbers when showing them, the format in the database is still readable.
On the member show page it does change the formatting though (+31 6 1234567890 instead of +3161234567890) so that's the reason I parsed it again there. Alternatively we could just skip the parsing and display the number without any formatting.

app/assets/javascripts/intro.js
FInnvos123: We can, no idea why I didn't do this

app/assets/stylesheets/intl_tel_input.css
FInnvos123: `intl_tel_input.css` now requires `intl-tel-input/build/css/intlTelInput`

db/migrate/20191030125101_normalize_phone_number.rb
```diff
@@ -0,0 +1,33 @@
...
+      m.update_attribute :phone_number, phone_number
+      m.update_attribute :emergency_phone_number, emergency_phone_number
+    end
+  end
+
+  def down
+    raise ActiveRecord::IrreversibleMigration
+  end
+
+  # normalize and validate phone number
```
maartenberg: This normalization logic probably isn't enough: we have a lot of phone numbers that are landlines, and some German and other odd values as well. I think that it's not a problem if we lose a member's phone number if it really is invalid, but we should in that case at least log the original value and which member the number belonged to (so the board can e.g. mail them if the member is still active).

Please try to handle Dutch local landlines: (`0[^0]\d{8}`, replace leading `0` with `+31`) which should also get the Dutch cellphones

FInnvos123: Landlines should now be handled.

Invalid numbers are added to the member's comments so we don't lose them.

db/migrate/20191030125101_normalize_phone_number.rb
```diff
@@ -0,0 +1,56 @@
+class NormalizePhoneNumber < ActiveRecord::Migration[5.2]
+  def up
+    Member.find_each do |m|
+      # duplicate numbers in case we need the old number
+      phone_number = normalize(m.phone_number.dup)
```
maartenberg: This can be a little shorter (without the `log_invalid_number`) by using the `<<` operator (append to sequences). Also, we only want to log the old emergency phone number if the member had one:

```suggestion
      new_phone_number = normalize(m.phone_number.dup)
      new_emergency_phone_number = normalize(m.emergency_phone_number.dup)
      
      comments = m.comments || ""
      comments << "\nMember had invalid phone number #{m.phone_number}" if new_phone_number.nil?
      comments << "\nMember had invalid emergency phone number #{m.emergency_phone_number}" if new_phone_number.nil? && m.emergency_phone_number.present?
```

PR #509, opened 2019-11-25
Add change password option for members

### LINKED ISSUE #393: Add change password to user edit member view
When a user is logged in he should be able to change his password right from the edit view, where he can edit his other information as well (address, etc). Is this (easily) done with devise?

FInnvos123:
Fixes #393 

Members can now change their password through the member edit page instead of having to click "forgot password".

### DIFF
--- a/app/views/user/password/edit.html.haml
+++ b/app/views/user/password/edit.html.haml
@@ -36,5 +36,6 @@
     .offset-xl-3.col-xl-6
       .card
         .card-body
-          %button.btn.btn-success{type: 'submit'} Submit
+          %button.btn.btn-success{type: 'submit'}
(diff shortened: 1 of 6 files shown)


app/views/user/password/edit.html.haml
```diff
@@ -0,0 +1,41 @@
...
+          %li= msg
+        - if flash[:error].present?
+          %li= flash[:error]
+
+  .row
+    .offset-xl-3.col-xl-6
+      .card
+        .card-header
+          %i.fa.fa-fw.fa-key
+          Wachtwoord veranderen
```
Riscky: This string is untranslated

PR #518, opened 2019-12-05
Improved products page

Mstiekema:
Fixed how the products page looked after the open source and added some extra functionality based on Bjorn his feedback.


app/assets/stylesheets/admin/checkout_products.css
```diff
@@ -30,6 +30,28 @@ table.table.table-linked > tbody > tr > td > a:hover {
   background: transparent;
 }
 
+.update-side-bar {
```
Riscky: The layout is unusable when on half screen. Both fullscreen and mobile work fine

Mstiekema: Fixed in new d9ebbc9

PR #520, opened 2019-12-05
Fixed incorrect yaml in the language files

Riscky:


Mstiekema:
e789448 shouldn't be in this PR

Issue #513, opened 2019-11-28
Remove search bar where unnecessary  


Mstiekema:
### Expected behaviour
Only display the search bar on the payment page & members page or other pages where necessary.

### Actual behavior
The search bar is displayed on a lot of pages where it serves no purpose.

### Steps to reproduce
- admin/user
- browser:

Riscky:
fixed in #555

PR #555, opened 2020-01-23
Removed search bar on admin pages that have no direct use for it

armaatus:
fixed issue #513


app/views/admin/activities/index.html.haml
```diff
@@ -2,7 +2,6 @@
 - content_for(:title, 'Activiteiten')
 
 %section#content
-  = render 'layouts/partials/search'
```
Riscky: I think this page has a use for the search bar

Riscky: My bad

PR #596, opened 2020-03-09
back button activity page

JustNireon:
Moved back button from bottom of activity page to left of the activity title


.rbenv-vars-sample
```diff
@@ -1,80 +0,0 @@
-### Database setup ###
```
Riscky: Don't delete this file please

PR #595, opened 2020-03-09
Icon updates

JustNireon:
Fixed #586 
Updated FA4 to FA5
This is with a new gem so files have been removed from /vendor.
Pull-left & pull-right have been refactored to float-left & float-right.

PR #597, opened 2020-03-09
external links

JustNireon:
Moved this commit away from #595 
Adds external links to the memberpage
Requires #595 to have the icons display


.rbenv-vars-sample
```diff
@@ -1,80 +0,0 @@
-### Database setup ###
```
Riscky: please don't remove this file

Issue #592, opened 2020-03-02
Add new Applied Data Science master


Riscky:
comparable to #477

PR #609, opened 2020-03-25
added ADS master, closing #592

Riscky:


Riscky:
uhm not really, I defs should do that

PR #610, opened 2020-03-28
remove non-bank supported characters from emails

Riscky:


Riscky:
There is something weird going on with the mail title which I want to look into first. Also I'd like the name to be consistent, which is not really the case right now: Activity title, mail subject and bank description are all in a different format.

PR #612, opened 2020-04-01
accept gif as poster, turn off resizing for gifs

Riscky:


app/views/admin/activities/partials/_edit.html.haml
```diff
@@ -29,8 +29,7 @@
           .col-md-4
             = label(:activity, :poster, class: 'd-none d-md-block')
             .div.poster.d-none.d-md-block
-              - if @activity.poster.attached?
```
Mstiekema, line 32: Removing this line breaks my edit activities page. Please bring it back. Everything else looks good.

Riscky: huh that is interesting, I'll check if I can reproduce the issue

PR #626, opened 2020-04-12
Bug/500 when removing activity

Riscky:
Old branch that could be rebased with ease.
It seems to work alright, but please check it thoroughly

### DIFF
--- a/app/models/participant.rb
+++ b/app/models/participant.rb
@@ -29,7 +29,7 @@ class Participant < ApplicationRecord
     impressions.each do |i|
       prefix = "#{ activity.name } (#{ activity.id }) - "
       message = case i.action_name
-                when "update" then
+                when "update"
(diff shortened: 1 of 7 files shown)


app/models/participant.rb
```diff
@@ -22,4 +23,21 @@ def currency
     self.paid = false if price_changed?
     write_attribute(:price, nil) if activity.price == self.price
   end
+
+  # Update logs before deleting a Participant to keep what happened
+  def rewrite_logs_before_delete!
```
Mstiekema, line 28: The result from this function is not as expected. The object now only shows the action (eg. "activities" for updating activity and "participants" when adding a participant) and the action is empty except when adding a user.
Screenshot from 2020-04-16 14-36-03

Riscky:
As far as I'm concerned we could merge this one already and look into the logging later (also see #377)

PR #631, opened 2020-04-13
Added download attribute to personal data download

### LINKED ISSUE #567: Redirect after personal info download
The member edit page can get 'stuck' on Firefox if the user downloads their personal information, as the url changes to https://leden.svsticky.nl/download and this breaks the Javascript somehow.
This is probably fixable by either opening the download in a new tab or redirecting the user back to the edit page after the download.

Mstiekema:
Fixed #567

### DIFF
--- a/config/locales/nl.yml
+++ b/config/locales/nl.yml
@@ -620,6 +620,7 @@ nl:
         edit_profile: Gegevens bewerken
         outstanding_payments: Openstaande betalingen
         upgrade: Opwaarderen
+        personal_data: "Persoonsgegevens"
     navigation:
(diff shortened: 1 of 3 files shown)


config/locales/nl.yml
```diff
@@ -620,6 +620,7 @@ nl:
         edit_profile: Gegevens bewerken
         outstanding_payments: Openstaande betalingen
         upgrade: Opwaarderen
+        personal_data: "Persoonsgegevens"
```
Riscky, line 623: can you add an english translation?

Mstiekema: Fixed it!

PR #638, opened 2020-04-17
Fully translated Koala

Mstiekema:
So it's time for this!

Current progress:
- [x] Added tool to track translations
- [x] Removed unused translations
- [x] Added missing translations (eg. en had it, nl did not have it yet)

Still have to do:
- [x] Find all the hardcoded strings in js / assests part of the application (requires the [i18n-js](https://github.com/fnando/i18n-js) gem)
	- [x] `lib/assets/javascripts/mail.js`
	- [x] `app/assets/javascripts/intro.js`
	- [x] `app/assets/javascripts/admin/*.js`
	- [x] `app/assets/javascripts/members/activities/*.js`
- [x] Find all the hardcoded strings in ruby / haml part of the application
	- [x] `app/controllers/api/checkout_controller.rb`
	- [x] `app/controllers/admin/checkout_products_controller.rb`
	- [x] `app/controllers/users/sessions_controller.rb`
	- [x] `app/mailers/mailings/*.rb`
	- [x] `app/views/admin/*`
	- [x] `app/views/layouts/partials/_search.html.haml`
	- [x] `app/views/members/*`
- [x] Split up translations in multiple files for better overview


Mstiekema:
I've look through every file except the main pile in `app/`. From what I could gather, the only file has to be translated there is `lib/assets/javascripts/mail.js`. These are just 3 simple string, but we would need the `i18n-js` gem to translate it. Not sure if that is worth it.

Mstiekema:
I've been working on the translations the past few hours and I noticed some duplicate code. Why do we have a seperate template for emails in `app/views/mailings/*`, but we sent out the templates in `app/mailers/mailings/*`? I would recommend using the first, unused templates. I am however not sure that it works this way. Can someone confirm this?

Mstiekema:
Finished translating everything! The mechanism to switch languages in user settings is not included. Will work on this if #506 remains dead. Language switcher should also include a way to switch the locale in the javascripts.

app/views/layouts/mailer.html.haml
```diff
@@ -61,7 +61,7 @@
             %table.main{:bgcolor => "#fff", :cellpadding => "0", :cellspacing => "0", :style => "font-family: 'Helvetica Neue',Helvetica,Arial,sans-serif; box-sizing: border-box; font-size: 14px; border-radius: 3px; background-color: #fff; margin: 0; border: 1px solid #e9e9e9;", :width => "100%"}
               %tr{:style => "font-family: 'Helvetica Neue',Helvetica,Arial,sans-serif; box-sizing: border-box; font-size: 14px; margin: 0;"}
                 %td.alert.alert-warning{:align => "center", :bgcolor => "#CE2029", :style => "font-family: 'Helvetica Neue',Helvetica,Arial,sans-serif; box-sizing: border-box; font-size: 16px; vertical-align: top; color: #fff; font-weight: 500; text-align: center; border-radius: 3px 3px 0 0; background-color: #CE2029; margin: 0; padding: 20px;", :valign => "top"}
-                  = I18n.t("association_name")
+                  = I18n.t('association_name')
```
Riscky, line 64: Maybe create a rule / test that checks this kind of formatting issues in CircleCI

app/views/members/activities/show.html.haml
```diff
@@ -33,8 +33,4 @@
             - @reservists.each do |participant|
               = render 'members/activities/partials/participant_table_row', participant: participant, activity: @activity
 
-= link_to "mailto:pleasedotheneedful@svsticky.nl" , class: "btn btn-danger helpme"do
```
Riscky, line 36: This is not really part of this PR, but I guess a rebase will solve the problem

spec/i18n_spec.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
```
Riscky, line 1: Check this spec in CircleCI

Mstiekema:
> 
> 
> Amazing work!
> Just a few comments
> Also I couldn't test the English translations yet, even though my browser language is set to English. I feel like this is an older issue, but I'd like to test before I approve the PR, so please reach out to help me change the language.

Thanks! At the moment, you can switch the language by adding the query parameter `?l=en` to the URL. As far as I know, there are still some issues with the javascript part, so I will be working on this later today / tomorrow. If you find any bugs there, let me know!

app/views/members/activities/show.html.haml
Mstiekema: Yep, fixed it!

Mstiekema:
I've rewritten the included Rspec to the testing model we use in this project. Before merging this PR, I would still like to write a proper implementation for the unused keys test. The `i18n-tasks health` tries to do this, but it catches some still used keys (The test does not catch the translations which use a predefined scope).

spec/i18n_spec.rb
Mstiekema: Fixed!

app/views/layouts/mailer.html.haml
Mstiekema: It's possible, but it would take a lot of time to fix throughout the application. I've only fixed a few in this PR, since they were giving me errors. So as far as I know, I think there might be a test for this for new translations, but not old ones

PR #643, opened 2020-04-22
Fix db:setup in local koala

Phavion:
Adds Redis to the docker-compose. db:setup needs this now, which caused the old version to fail as redis was not present.


fiinnnn:
LGTM

Issue #336, opened 2018-03-09
DRY mailing views


cdfa:
The mailing view contain a lot of duplicated code for the table layouts (especially the inline styles). This has a bad impact on readability and maintainablity. We should DRY it up a little.

PR #647, opened 2020-04-29
Multilingual Koala

### LINKED ISSUE #486: Make Koala multilingual
# Things that need to be done:
- [x] Update all missing translations (related: #396, #327)
- [x] Make a language choice thingy (flag in the edge of the page)
- [x] Make activity descriptions bilingual (optional) (tabs?)
- [x] Display all pages and content in the chosen language
- [x] Save user language preference

Mstiekema:
The settings page now features a selector in which you can select your preferred language. The activity description is only shown in your preferred language as well. Fixes #486

### DIFF
--- a/app/assets/javascripts/members/activities/activity.js
+++ b/app/assets/javascripts/members/activities/activity.js
@@ -1,6 +1,4 @@
 //= require cache_helpers.js
-//= require i18n
-//= require i18n/translations
 //= require language.js
 
 /**
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -27,13 +27,9 @@ class Activity < ApplicationRecord
 
   validates :notes, presence: true, if: proc { |a| a.notes_public? || a.notes_mandatory? }
 
-  # Disabled validations
-  # validates :end_date
-  # validates :unenroll_date
+  is_impressionable
 
   before_destroy :rewrite_logs_before_delete, prepend: true
-  is_impressionable :dependent => :ignore
-
   after_update :enroll_reservists!, if: proc { |a| a.saved_change_to_participant_limit }
 
   has_one_attached :poster
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -29,7 +29,6 @@ class Activity < ApplicationRecord
 
   # Disabled validations
   # validates :end_date
-  # validates :description
   # validates :unenroll_date
 
   before_destroy :rewrite_logs_before_delete, prepend: true
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -3,6 +3,8 @@ class User < ApplicationRecord
   is_impressionable
 
   belongs_to :credentials, :polymorphic => true
+  
+  enum language: [:nl, :en]
 
   # Include default devise modules. Others available are:
   # :lockable, :timeoutable and :omniauthable
@@ -13,10 +15,6 @@ class User < ApplicationRecord
     return true if credentials_type.casecmp('admin').zero?
 
     return false
-  end
-
-  def language
-    return credentials.language
   end
 
   def sender
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -15,6 +15,10 @@ class User < ApplicationRecord
     return false
   end
 
+  def language
+    return credentials.language
+  end
+
   def sender
     "#{ credentials.name } <#{ email }>"
   end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -3,7 +3,7 @@ class User < ApplicationRecord
   is_impressionable
 
   belongs_to :credentials, :polymorphic => true
-  
+
   enum language: [:nl, :en]
 
   # Include default devise modules. Others available are:
--- a/app/views/admin/settings/index.html.haml
+++ b/app/views/admin/settings/index.html.haml
@@ -103,9 +103,10 @@
 
               .form-group
                 .row
-                  %label{for: ''}
+                  .col-md-12
+                    %label{for: ''}
                     = I18n.t('activerecord.attributes.member.language')
-                    = f.select :language, options_for_select([["Nederlands", :nl], "Engels"], selected: @user.language)
+                    = f.select :language, options_for_select([["Nederlands", :nl], ["Engels", :en]], selected: @user.language)
 
               .form-group#mail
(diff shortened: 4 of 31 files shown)


app/assets/javascripts/members/activities/activity.js
```diff
@@ -1,6 +1,4 @@
 //= require cache_helpers.js
-//= require i18n
```
Riscky, line 2: This doesn't seem right, the file does contain i18n.t statements

app/models/activity.rb
```diff
@@ -29,7 +29,6 @@ def content_type
 
   # Disabled validations
   # validates :end_date
-  # validates :description
```
Riscky, line 32: Why should this comment be removed?

app/models/user.rb
```diff
@@ -4,6 +4,8 @@ class User < ApplicationRecord
 
   belongs_to :credentials, :polymorphic => true
 
+  enum language: [:nl, :en]
```
Riscky, line 7: Why is the enum present in both user and member?

app/views/admin/settings/index.html.haml
```diff
@@ -101,6 +101,13 @@
                     = I18n.t('admin.members.surname')
                     = f.text_field :last_name, :value => @admin.last_name, :class => 'form-control'
 
+              .form-group
+                .row
+                  .col-md-12
+                    %label{for: ''}
+                    = I18n.t('activerecord.attributes.member.language')
+                    = f.select :language, options_for_select([["Nederlands", :nl], ["Engels", :en]], selected: @user.language)
```
Riscky, line 109: Shouldn't those string be translated as well (Dutch, English)?

app/views/api/activities/index.rabl
```diff
@@ -1,6 +1,6 @@
 collection @activities
 
-attribute :id, :name, :location, :price, :show_on_website, :description
+attribute :id, :name, :location, :price, :show_on_website, :description_nl, :description_en
```
Riscky, line 3: Probably the activity section on the site won't work after this change, we should defs have PR ready in static-sticky

db/migrate/20200429072750_multilingual_activity_description.rb
```diff
@@ -0,0 +1,6 @@
+class MultilingualActivityDescription < ActiveRecord::Migration[6.0]
+  def change
+    rename_column :activities, :description, :description_nl
```
Riscky, line 3: nice

db/seeds/activities.rb
```diff
@@ -38,7 +38,8 @@
 
     location:          Faker::TvShows::FamilyGuy.location,
     organized_by:      Faker::Boolean.boolean(true_ratio: 0.8) ? Group.all.sample : nil,
-    description:       Faker::Lorem.paragraph(sentence_count: 5),
+    description_nl:    Faker::Lorem.paragraph(sentence_count: 5),
```
Riscky, line 41: Faker can generate multilingual texts afaik

db/seeds/members.rb
```diff
@@ -18,7 +18,8 @@
     student_id:   "F#{ Faker::Number.number(digits: 6) }",
     birth_date:   Faker::Date.between(from: 28.years.ago, to: 16.years.ago),
     join_date:    Faker::Date.between(from: 6.years.ago, to: Date.today),
-    comments:     (Faker::Boolean.boolean(true_ratio: 0.3) ? Faker::Hacker.say_something_smart : nil)
+    comments:     (Faker::Boolean.boolean(true_ratio: 0.3) ? Faker::Hacker.say_something_smart : nil),
+    language:     0
```
Riscky, line 22: maybe set language randomly?

Riscky:
image
This looks weird, my suggestion would be to move the birthdate above the warning field anyway.

app/views/admin/settings/index.html.haml
Mstiekema: This was something we discussed when implementing the language switch in `static-sticky` as well. Most of the times when switching languages, you would want to see the text in its native language. This is pretty much standard everywhere.

Riscky:
image
Also check the mobile UI

app/models/activity.rb
Mstiekema: Forgot to replace it there, but I also think this comment section is no longer needed since it has not been touched in the last 3 years

app/views/admin/settings/index.html.haml
Riscky: In that case, shouldn't it be Nederlands, English?

Mstiekema: Oh shoot, you're right, will fix that whenever I have time

app/models/activity.rb
Riscky: In that case just remove the whole block

app/views/api/activities/index.rabl
Mstiekema: Did that! Available at https://github.com/svsticky/static-sticky/pull/236

app/models/user.rb
Mstiekema: This was a leftover from previous code, was removed

db/seeds/activities.rb
Mstiekema: So I've tried a lot of different Faker types and in the end I found the [NL locale file](https://github.com/faker-ruby/faker/blob/master/lib/locales/nl.yml). It turns out they do not have a lot of Dutch translations, so I think we shoudl stick to the Lorem text

db/migrate/20200429072750_multilingual_activity_description.rb
Mstiekema: ty :)

Mstiekema:
Improved CSS for desktop and mobile, so everything should be fixed now

Mstiekema:
Important note: Admins _can_ change language, but it is not suggested at the moment, because you will send out emails, whatsapp messages, etc. in the preferred language. 
Maybe it would be a good idea to open an issue to send out these messages in users their preferred language? Would also combine this issue with #336 to clean up the entire mailing of the application

db/seeds/activities.rb
Riscky: makes sense, thanks for looking into it

PR #665, opened 2020-05-06
Recent products

Mstiekema:
Also has some minor improvements for the card info. This PR is mainly focussed on the new implementation of Checkout, [magnificent-sloth](https://www.github.com/svsticky/magnificent-sloth).


Mstiekema:
Only show alcohol products after set time limit

JustNireon (approved):
LGTM

PR #673, opened 2020-05-13
fixed bug + set fallback description

Riscky:


app/views/members/activities/partials/_view.html.haml
```diff
@@ -71,6 +71,10 @@
         = simple_format(activity.description_nl)
       - elsif @user.language == 'en' && !activity.description_en.blank?
         = simple_format(activity.description_en)
+      - elsif !activy.description_nl.blank?
+        = simple_format(activity.description_nl)
+      - elsif !activity.description_en.black?
+        = = simple_format(activity.description_en)
```
Mstiekema, line 77: double '='

PR #687, opened 2020-05-22
display comma-separated committee roles from previous years

Riscky:


fiinnnn:
The current role could be highlighted, the rest lgtm

PR #688, opened 2020-05-22
fixed error where members without an account could not be shown or edited

Riscky:


app/views/admin/members/edit.html.haml
```diff
@@ -123,7 +123,7 @@
                         %span.fa.fa-trash
 
 
-                - unless ENV['MAILCHIMP_DATACENTER'].blank?
+                - unless ENV['MAILCHIMP_DATACENTER'].blank? || @member.user.nil?
```
fiinnnn, line 126: The error seems to happen because mailchimp_interests returns nil.
This happens when the email address isn't on the mailing list.
Instead the mailchimp interests could be hidden if `@member.mailchimp_interests.nil?`

app/views/admin/members/show.html.haml
```diff
@@ -97,7 +97,7 @@
                     %span.label.label-info
                       = I18n.t('activerecord.attributes.education.status.active')
 
-            - unless ENV['MAILCHIMP_DATACENTER'].blank?
+            - unless ENV['MAILCHIMP_DATACENTER'].blank? || @member.user.nil?
```
fiinnnn, line 100: Same here

PR #693, opened 2020-05-22
fix intro & status page locale switch

Riscky:


app/assets/javascripts/language.js
```diff
@@ -7,7 +7,7 @@ I18n.fallbacks = true;
 I18n.defaultLocale = "nl";
 
 // Language switcher
-const urlParams = new URLSearchParams(window.location.search);
```
Mstiekema, line 10: Why did you change this? Both are constants

Mstiekema:
"Gezelligheid" is not translated either

app/assets/javascripts/language.js
Riscky: it gave errors in the console on switching language

Riscky:
> "Gezelligheid" is not translated either

There is no translation for gezelligheid, people should learn at least this one dutch word ;)

app/assets/javascripts/language.js
Mstiekema: Hmm weird, okay!

PR #705, opened 2020-06-07
update group.js

JustNireon:
Minor fix which makes it possible to remove group members

### DIFF
--- a/app/assets/javascripts/admin/apps.js
+++ b/app/assets/javascripts/admin/apps.js
@@ -147,7 +147,7 @@ $(document).on( 'ready page:load turbolinks:load', function(){
     var token = encodeURIComponent($(this).closest('.page').attr('data-authenticity-token'));
     var row = $(this).closest('tr');
 
-    if( !confirm(I18n.t('admin.general.remove', $(row).find('a').html())) )
+    if( !confirm(I18n.t('admin.general.remove', {user:$(row).find('a').html()})) )
       return;
 
     $.ajax({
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/admin/apps.js
```diff
@@ -147,7 +147,7 @@ $(document).on( 'ready page:load turbolinks:load', function(){
     var token = encodeURIComponent($(this).closest('.page').attr('data-authenticity-token'));
     var row = $(this).closest('tr');
 
-    if( !confirm(I18n.t('admin.general.remove', $(row).find('a').html())) )
```
Riscky, line 150: I get: `SyntaxError: missing } after property list`
(on Firefox)
or `Uncaught SyntaxError: Unexpected token ')'`
(on Chromium)

Riscky: op whatever pagina aan de adminkant trouwens

app/assets/javascripts/admin/groups.js
```diff
@@ -14,7 +14,7 @@ function bind_group_members(){
     var token = encodeURIComponent($(this).closest('.page').attr('data-authenticity-token'));
     var row = $(this).closest('tr');
 
-    if( !confirm(I18n.t('admin.general.remove', $(row).find('a').html())) )
+    if( !confirm(I18n.t('admin.general.remove', {user:$(row).find('a').html()))} )
```
Riscky, line 17: `    if( !confirm(I18n.t('admin.general.remove', {user:$(row).find('a').html()})) )`

^ this should solve the problem

PR #707, opened 2020-06-07
Added Pagy for better pagination

Mstiekema:


Gemfile
```diff
@@ -10,6 +10,7 @@ gem 'rabl'
 # assets and stuff
 gem 'coffee-rails'
 gem 'font_awesome5_rails'
+gem 'pagy'
```
Riscky, line 13: shouldn't will_paginate and will_paginate-bootstrap4 be removed too?

Riscky: Still need to update Gemfile.lock

PR #740, opened 2020-09-01
New login page based on Tailwind

Riscky:
I recently became a big fan of Tailwind because it gives me the control over the layout without ever leaving the html.

Tailwind needs webpack, so this PR is based off #775 

I copied our current color and breakpoint configuration to Tailwind to ease working with both Yeti and Tailwind

I merged the example login page into this branch, as it is as good as finished


app/views/layouts/doorkeeper.html.haml
```diff
@@ -1,4 +1,6 @@
 !!! 5
+= javascript_pack_tag 'application'
```
Riscky, line 2: this changes are only for testing purposes and should be removed before merge

package.json
```diff
@@ -9,14 +9,25 @@
     "Maarten van den Berg <>"
   ],
   "dependencies": {
+    "@fullhuman/postcss-purgecss": "^2.3.0",
+    "@rails/webpacker": "5.2.1",
     "bootstrap": "^4.3.1",
     "clipboard": "^1.6.1",
+    "core-js": "^3.6.5",
```
Riscky, line 16: I'm not actually sure what each of this packages do, so I'll have to clean this bit out

config/webpack/development.js
```diff
@@ -0,0 +1,5 @@
+process.env.NODE_ENV = process.env.NODE_ENV || 'development'
+
+const environment = require('./environment')
```
Riscky, line 3: this file is currently not run automatically on server start, which is annoying

config/webpacker.yml
```diff
@@ -0,0 +1,92 @@
+# Note: You must restart bin/webpack-dev-server for changes to take effect
```
Riscky, line 1: auto generated, but probably alright?

tailwind.config.js
```diff
@@ -0,0 +1,47 @@
...
+      success: '#43ac6a',
+      info: '#5bc0d',
+      warning: '#E99002',
+      danger: '#F04124',
+      light: '#eee',
+      dark: '#222',
+    },
+    screens: {
+      // this breakpoints are copied from Yeti
+      'xs': '0',
```
Riscky, line 38: I might want to remove this breakpoint, since it is not very efficient to generate css for a breakpoint like this XD

postcss.config.js
```diff
@@ -0,0 +1,28 @@
+let environment = {
+  plugins: [
+    require('tailwindcss'),
+    require('autoprefixer'),
+    require('postcss-import'),
```
Riscky, line 5: same here, there might be steps here that don't make sense for us

tailwind.config.js
```diff
@@ -0,0 +1,46 @@
+
+module.exports = {
+  purge: {
+    mode: 'production',
+    content: ['./src/**/*.html'],
+  },
+  theme: {
+    extend: {},
+    colors: {
+      // this colors are copied from Yeti
```
Mstiekema, line 10: These*

tailwind.config.js
```diff
@@ -0,0 +1,46 @@
+
+module.exports = {
+  purge: {
+    mode: 'production',
+    content: ['./src/**/*.html'],
+  },
+  theme: {
+    extend: {},
+    colors: {
```
Mstiekema, line 9: Isn't this a perfect spot to add the board color?

tailwind.config.js
```diff
@@ -0,0 +1,46 @@
...
+      secondary: '#eee',
+      success: '#43ac6a',
+      info: '#5bc0d',
+      warning: '#E99002',
+      danger: '#F04124',
+      light: '#eee',
+      dark: '#222',
+    },
+    screens: {
+      // this breakpoints are copied from Yeti
```
Mstiekema, line 35: These*

Mstiekema:
Looks neat btw! Did however had to run `bundle exec rails webpacker:install` for some reason. Not sure if I did something wrong if it's actually required :/

tailwind.config.js
Riscky: I guess it is, but then we could also not include a whole lot of the other colors and use a consistent color scheme instead. That feels a bit beyond the scope of this PR though.

Riscky:
> Looks neat btw! Did however had to run `bundle exec rails webpacker:install` for some reason. Not sure if I did something wrong if it's actually required :/

This should defs not be necessary, but I will wait for #776 to be merged before diving into this one. #776 might actually fix build-related problems

Mstiekema (approved):
LGTM! :shipit:

PR #747, opened 2020-10-01
Fix checkout product sorting

### LINKED ISSUE #719: checkout product sorting
### Expected behaviour
Products are sorted if they are active and afterwards alphabetically

### Actual behavior
When a product gets selected it only gets sorted alphabetically but not if the product is active

### Steps to reproduce
- admin
1. select product

fiinnnn:
Closes #719


JustNireon (approved):
LGTM

PR #752, opened 2020-10-02
Fixed active link

### LINKED ISSUE #749: Sidebar item doesn't stay highlighted
On the members page (page is `/members`):
image

After clicking a member (so the page is `/members/id`):
image

Currently an item is highlighted based on the output of `current_page?`, this means `/members` and `/members/id` are treated as two different things despite being in the same category on the sidebar.

The same thing happens on the products page after selecting a product.

JustNireon:
closes #749 
Kinda baffled by this bug. I'm not sure if this is our fault or something else but this fixes the issue

### DIFF
--- a/app/views/layouts/partials/_navigation.html.haml
+++ b/app/views/layouts/partials/_navigation.html.haml
@@ -7,7 +7,7 @@
 
       = link_to :members, :class => 'list-group-item list-group-item-action' do
         %span= I18n.t("admin.navigation.member")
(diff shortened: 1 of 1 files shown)


fiinnnn:
LGTM, could the same fix be applied to the checkout products page?
The same thing happens there when selecting a product.

JustNireon:
I've fixed it for products it seems like it's for all pages that have a index page and a show page so i did it for all of them.

fiinnnn (approved):
LGTM

PR #760, opened 2020-10-08
Fix participant table update after enrolling

### LINKED ISSUE #454: Enrolling for an activity doesn't immediately add the member's name to the list
### Expected behaviour
- Enroll for activity as a member
- Participant count in activity card increases
- Participant count above participant list increases
- Participant name is added to the participant list

### Actual behavior
- Participant count in card increases
- Participant list and count above do not change

### Steps to reproduce
- Detail view of any activity (tested with 759, 760)
- Enroll

Tested on Firefox v66.

fiinnnn:
Closes #454 

This should probably be refactored at some point.
For now I restored the old authorize function `api/activities_controller` and the string formatting function in `members/main.js`.

### DIFF
--- a/app/assets/javascripts/members/main.js
+++ b/app/assets/javascripts/members/main.js
@@ -1,6 +1,14 @@
 $(document).on("ready page:load turbolinks:load", function () {
   setup_intl_tel_input();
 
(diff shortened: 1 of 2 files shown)


Mstiekema (approved):
LGTM

PR #770, opened 2020-10-16
Fixed calendar copy link

Mstiekema:
I'm not even sure which applications use webcal as protocol. As far as I've seen using https is best practice and just gets you the .ics file :)


Riscky:
> I'm not even sure which applications use webcal as protocol. As far as I've seen using https is best practice and just gets you the .ics file :)

nobody want the .ics file though, because that does not give you updates. webcal is supported by google calender and ms outlook at least. I don't _like_ the current solution, but changing the link back to http isn't going to help anyone.

PR #775, opened 2020-11-12
Packages/webpack

Riscky:
TODO:

- [x] Make sure CSRF token is sent on javascript forms (jquery-ujs doesn't seem to work)
- [x] Clean up packages.json, environment.js
- [x] Make sure prettier is configured to check the new folders in the CI
- [x] Fix signout (the 'delete' request is changed to post for some reason)
- [x] Move inline js of admin/settings/index.html.haml to script + import $ (fixes the last two issues of Merijn)
- [x] Fix long compile times and large bundle sizes
- [x] Fix signing up 
- [x] Fix updating info for members
- [x] Removing user from activity returns Uncaught TypeError: a.magic_reservists is undefined (probably only with non-full activities?)
- [x] Rebase
- [x] Webpack in nix-shell
- [x] Nix build fails: rails-erb-loader
- [x] Tested on staging
- [x] Live reloads with webpack-dev-server :) (optional)

I used a CDN for the phonenumber lib (which ís lazy loaded), but we might actually want to use the included lib. This can be another issue though.
We might want an issue to move our css to webpack in order to be able to remove Sprockets completely (but this is not default in rails 6: https://rossta.net/blog/why-does-rails-install-both-webpacker-and-sprockets.html)
A more general remark: we should defs clean up our javascript in the future, it's a bit of a mess.


Mstiekema:
Tested a few things, this is what I encountered:
- Can't sign up for activities or updating extra info, `/activities` & `/activities/:id` both give the error in the console `Activity is not defined`.
- Updating roles for people in committees sometimes does something weird / doesn't give you the correct role.
- Removing user from activity from their user page seems to work, but returns `Uncaught TypeError: a.magic_reservists is undefined` in the console.
- As an admin, going to `/settings` returns `Uncaught ReferenceError: $ is not defined`
- Updating the signature fails and returns `Assertion failed: Input argument is not an HTMLInputElement` in the console.

Riscky:
`Updating roles for people in committees sometimes does something weird / doesn't give you the correct role.`
This bug is also present in the current development branch, so I think we need to look into that in a separate issue

Mstiekema:
> This bug is also present in the current development branch, so I think we need to look into that in a separate issue.
> Update: made #780
Great, thanks!

Mstiekema (approved):
Tested everything I could. Previous bugs I encountered have been fixed. LGTM! :shipit:

Mstiekema:
Very cool, very nice. Everything works as intended. Please add a notice in the README to remove `node_modules` if still present. If this is fixed we can ship it :shipit:

Mstiekema (dismissed):

PR #776, opened 2020-11-12
Nix build

Riscky:
This PR provides us with a complete Nix build, which includes ditching Yarn.
This should aid server installations

TODO
- [x] log directory is not writable
- [x] update README, INSTALLING and DEPENDENCIES to reflect changes
- [x] fix dependency on directory name (only if there's an easy fix)

tmp directory is not writable, but this is not fixable, unless we include a systemd file


Mstiekema (dismissed):
Reviewed & tested, LGTM!

PR #784, opened 2020-12-10
Update email address in mailchimp when a member's email is changed

### LINKED ISSUE #779: Email address isn't updated in mailchimp
### Expected behaviour
When a user changes their email in koala the email should be updated in mailchimp as well

### Actual behavior
In mailchimp the address is never changed, so afterwards it seems like the user doesn't exist in mailchimp.
This means they can no longer change their mailing preferences.

Related to #715 and #700

fiinnnn:
Closes #779


Mstiekema (approved):
Looks great! :shipit:

PR #792, opened 2021-01-14
Email change enhancements

### LINKED ISSUE #715: Changing email address
Requirements:
- [x] New email address has to be confirmed (making sure it exists and the member has access to the email address)
- [x] Member should be warned on every page that the confirmation is pending
- [ ] Member should login before confirmation succeeds (making sure hostile account takeover is not possible without knowing the password). Probably using the new email address?
- [x] Mailchimp preferences should be updated to the new account (removing the old email address and creating the new one with the right preferences if the api does not support changing an email address)

Related to #700

fiinnnn:
- [x] Member should be warned on every page that the confirmation is pending
- [x] Member should login before confirmation succeeds

Closes #715

### DIFF
--- a/app/controllers/members/home_controller.rb
+++ b/app/controllers/members/home_controller.rb
@@ -77,9 +77,6 @@ class Members::HomeController < ApplicationController
 
       cookies["locale"] = @user.language
 
-      # the translation location was used here but that conflicted with the way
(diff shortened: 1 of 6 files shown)


Riscky:
dont crash on missing mailchimp credentials

Riscky (approved):
checked and works :)

PR #794, opened 2021-01-14
Add development postgres support, fix hardcoded queries

Phavion:
This pull request adds support for postgres on koala.

-- config/database.yml still needs adaptation for when this is to be run on staging or production!
-- iow, still needs some work.


Mstiekema (approved):
Let's go! :shipit:

PR #797, opened 2021-02-12
Fix dropdown highlights for group members

JustNireon:
Fixes minor issue that when searching members in the group page they don't get highlighted

### DIFF
--- a/app/assets/stylesheets/admin/groups.css
+++ b/app/assets/stylesheets/admin/groups.css
@@ -54,3 +54,9 @@ table.table#members tr td.buttons button {
   top: 4px;
   margin-right: 2px;
 }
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/admin/groups.css
```diff
@@ -54,3 +54,7 @@ table.table#members tr td.buttons button {
   top: 4px;
   margin-right: 2px;
 }
+li.active {
+    color: #fff;
+    background: #008cba;
+}
```
Mstiekema, line 60: Add `cursor: pointer` as property as well

Mstiekema (changes requested):

PR #805, opened 2021-03-09
Disable NFC cards

### LINKED ISSUE #453: Disable activated NFC cards
Currently under Apps -> Checkout (in the board member view) we can activate unactivated NFC cards. Some people ask to disable cards they do not use anymore, or there card is lost etc.

We would like to see an option to remove a card from any user. Only board members should be able to do this.

CrookedGrandma:
Closes #453 

Adds a new table (card) to the member overview, in which the member's cards can be seen. 
For each card, the UUID is shown and its status next to that.

From this overview, you can:
- Activate cards that have not been activated yet
- Deactivate cards
- Reactivate deactivated cards

### DIFF
--- a/app/assets/javascripts/admin/members.js
+++ b/app/assets/javascripts/admin/members.js
@@ -157,29 +157,61 @@ $(document).on("ready page:load turbolinks:load", function () {
 
   destroy(null);
 
-  $("#cards button.deactivate").on("click", deactivate);
+  $("#cards button").on("click", button_action);
 
-  function deactivate() {
+  function button_action() {
     let row = $(this).closest("tr");
+    let status = $(row).children('.cardstatus');
     let uuid = row.attr("data-uuid");
     let memberid = row.attr("data-member-id");
     let token = encodeURIComponent(
       $(this).closest(".page").attr("data-authenticity-token")
     );
 
-    if (!confirm(
-      I18n.t("admin.cards.deactivate_confirm", { uuid: uuid })
-    )) return;
+    let tobeactivated = row.attr("data-active") == 0;
+    let disabled = row.attr("data-disabled") == 1;
+
+    let entry = "admin.cards." + (tobeactivated ? "" : (disabled ? "re" : "de")) + "activate_confirm";
+    if (!confirm(I18n.t(entry, { uuid: uuid }))) return;
+
+    let url = tobeactivated ? "/apps/card"
+      : "/members/" + memberid + "/set_card_disabled/" + uuid;
+    let successmsg = tobeactivated ? I18n.t("checkout.card.activated")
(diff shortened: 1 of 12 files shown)


app/controllers/api/checkout_controller.rb
```diff
@@ -133,4 +124,12 @@ def authenticate_checkout
       nil
     end
   end
+
+  def authenticate_card
+    @uuid = params[:uuid]
+    @card = CheckoutCard.find_by(uuid: @uuid)
```
JustNireon, line 130: Should first check if card exists otherwise return 404. @card.active gives an error now since @card is nil in certain scenarios

app/assets/javascripts/admin/members.js
```diff
@@ -156,4 +156,85 @@ $(document).on("ready page:load turbolinks:load", function () {
   });
 
   destroy(null);
+
+  $("#cards button").on("click", button_action);
+
+  function button_action() {
```
JustNireon, line 162: This function also get's called in admin/checkout when activating / removing cards not really sure how but it happens

app/controllers/api/checkout_controller.rb
```diff
@@ -133,4 +124,12 @@ def authenticate_checkout
...
+    render status: :unauthorized, json: I18n.t('checkout.error.not_activated') unless @card.active
```
JustNireon, line 131: Should check if @card is nil before this now it happens afterwards in 'info' or never

JustNireon (changes requested):

app/controllers/api/checkout_controller.rb
```diff
@@ -133,4 +124,13 @@ def authenticate_checkout
       nil
     end
   end
+
+  def authenticate_card
+    render status: :not_found if @card.nil?
```
JustNireon, line 129:
```suggestion
    
```

app/controllers/api/checkout_controller.rb
```diff
@@ -133,4 +124,13 @@ def authenticate_checkout
...
+    @uuid = params[:uuid]
+    @card = CheckoutCard.find_by(uuid: @uuid)
+    render status: :unauthorized, json: I18n.t('checkout.error.not_activated') unless @card.active
```
JustNireon, line 132:
```suggestion
    render status: :not_found && return if @card.nil?
    render status: :unauthorized, json: I18n.t('checkout.error.not_activated') unless @card.active
```

JustNireon (changes requested):
First have to initiate card to check if it's nil :)

app/controllers/api/checkout_controller.rb
CrookedGrandma: Lmao literally -69 IQ, fixed

JustNireon (approved):
LGTM

PR #808, opened 2021-03-23
Activity summary views

### LINKED ISSUE #434: Export participants of activity
We thought it would be a useful to be able to export (to a simple .csv or similar) participants of an activity. Often we need a list of participants on paper for something during the activity (keep track of what drinks were bought, who is present etc.), sometimes we'd even like to simply export their e-mails so we can manually send mails or similar.

For this we'd like to be able to select what columns to export. Columns we thought of are `Naam`, `Reserve` (a boolean if the participant is on the spares list or not), `Prijs`, `Nog betalen` (a boolean if the participant still needs to pay or not), `Extra opmerking`, `E-mail`. With option to add empty columns with a custom name.

Raymonzut:
Closes #434

Introduces richer views for activity-participants

### DIFF
--- a/app/assets/javascripts/admin/activities.js
+++ b/app/assets/javascripts/admin/activities.js
@@ -417,6 +417,8 @@ function formatTableAsCSV() {
       .last()
       .siblings()
       .each(function (__) {
+        if ($(this).text().length <= 1)
+          $(this).addClass("activity_table-cell--pseudo-hidden");
         $(this).append(",");
       });
   });
--- a/app/assets/javascripts/admin/activities.js
+++ b/app/assets/javascripts/admin/activities.js
@@ -370,11 +370,14 @@ $(document).on("ready page:load turbolinks:load", function () {
     });
 
   posterHandlers();
-  if (window.location.href.indexOf("summary_only") !== -1)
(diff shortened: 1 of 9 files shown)


JustNireon (changes requested):
Pressing show as csv twice adds extra lines to notes which is kinda weird. furthermore we yarn has a package that makes it real easy to add a "Copy to Clipboard" Button it would be a pretty nice feature to have aswell

Raymonzut:
@JustNireon Alright, done, anything else for this feature?

JustNireon:
Ill check copy to.clipboard this afternoon but otherwise lgtm

JustNireon:
When pressing show as CSV when it's already summarized it still looks weird like this: 
image
Furthermore for copying to clipboard i would use the similar code used in '/app/assets/javascripts/admin' for consistency. 
`  (function () {
    new Clipboard("#copy_transactions .btn-clipboard");
  })();`

`      // Bind json data to copy button
      $("#copy_transactions button").attr(
        "data-clipboard-text",
        JSON.stringify(data)
      );`

JustNireon (approved):
LGTM

PR #822, opened 2021-05-26
Query for non-existent user after destroy

JustNireon:
When the user deletes their account the impressionist page tries to query the user but fails and makes the entire page fail. This should fix it.


Mstiekema (approved):
lgtm

PR #823, opened 2021-05-28
Fixes imppresionist view 

JustNireon:
Actually fixes the issue that #822 should've fixed. Probably should've tested this and not made this PR entirely on github.com


Mstiekema (approved):
tested, lgtm

PR #825, opened 2021-06-08
Payconiq staging issues

JustNireon:
- mongoose returns array instead of hash in export csv
- symbol :successful instead of 'SUCCEEDED' 
- Make activity payment without any activities
- Removed is_payable_before_unenroll_date validation since this should be up to the treasurer not the system
- Updated sample.env
- minor issues


Mstiekema (approved):
lgtm

PR #826, opened 2021-06-15
Fix seeding & Search & Transaction accountancy

JustNireon:
Some minor fixes before deploying payconiq onto production


Mstiekema (approved):
lgtm

PR #827, opened 2021-06-15
Fix intro site and minor translation

JustNireon:
Intro site still had old payment system. Has been updated to new payment model with only ideal for now

### DIFF
--- a/app/controllers/admin/payments_controller.rb
+++ b/app/controllers/admin/payments_controller.rb
@@ -52,10 +52,11 @@ class Admin::PaymentsController < ApplicationController
   end
 
   def export_payments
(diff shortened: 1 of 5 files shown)


app/controllers/admin/payments_controller.rb
```diff
@@ -52,10 +52,11 @@ def update_transactions
   end
 
   def export_payments
+    payment_type = params[:payment_type] == "Payconiq" ? [:payconiq_online, :payconiq_display] : [:ideal]
```
Mstiekema, line 55: Why do we give an array with both payconiq types? Shouldn't we only have the online one?

JustNireon: All payconiq transactions currently are paid in one transaction but this is irrelevant for now as long as the payconiq display has no functionality

Mstiekema (approved):
lgtm

PR #859, opened 2021-07-28
Darktheme login

Riscky:
This completely removes Bootstrap from the Doorkeeper pages, replacing everything with Tailwind.
Also, it adds a beautiful dark theme for this pages.
I'm not a designer, suggestions for improvements are welcome.

There is a bit of duplication in the code, but it has been there for a while, so I'm not fixing it in this PR.
Someone should DRY up the different Doorkeeper / Devise screens however.

I removed the Yeti colors in favor of the default Tailwind colors.
This set of colors is just more complete (and I don't really care about Yeti's feelings).

Also included is a bit of Webpack config changes that I came across while developing.

TODO: 
- [ ] Test the pages thoroughly, there are A LOT of different screens and warnings and I have not checked every single one of them.
- [x] Fix subtitle color (for example in /password/new)
- [x] Use a non-black background

This is how the new dark theme looks like:
image

### DIFF
--- a/app/views/layouts/doorkeeper.html.haml
+++ b/app/views/layouts/doorkeeper.html.haml
@@ -22,7 +22,7 @@
       .shadow.px-10.py-4.rounded-md.bg-white.mx-auto
 
         - if current_user.present?
-          %h2
+          %h2.mt-2
             =I18n.t('layouts.doorkeeper.hi')
             %b= current_user.credentials.first_name
             = I18n.t('layouts.doorkeeper.identity_confirmation', name: current_user.credentials.name)
--- a/app/views/layouts/doorkeeper.html.haml
+++ b/app/views/layouts/doorkeeper.html.haml
@@ -15,11 +15,12 @@
     = javascript_pack_tag 'doorkeeper', 'data-turbolinks-track': 'reload'
 
     = csrf_meta_tags
-  %body.h-full.flex.flex-col.bg-gray-100
+  %body.h-full.flex.flex-col.bg-gray-100.dark:bg-black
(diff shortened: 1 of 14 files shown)


app/views/layouts/doorkeeper.html.haml
```diff
@@ -15,14 +15,15 @@
...
     = csrf_meta_tags
-  %body.h-full.flex.flex-col.bg-light
+  %body.h-full.flex.flex-col.bg-gray-100.dark:bg-gray-900
     .flex.flex-col.flex-grow.justify-center
-      %img.w-64.mx-auto.p-4{src:"https://public.svsticky.nl/logos/logo_compact_kleur.svg"}
+      %img.w-64.mx-auto.p-4.dark:hidden{src:"https://public.svsticky.nl/logos/logo_compact_kleur.svg"}
+      %img.w-64.mx-auto.p-4.hidden.dark:block{src:"https://public.svsticky.nl/logos/logo_compact_outline_wit.svg"}
 
-      .shadow.px-10.py-4.rounded-md.bg-white.mx-auto
+      .shadow.px-10.py-4.rounded-md.bg-white.mx-auto.dark:bg-gray-800.dark:shadow-
```
JustNireon, line 23: Shouldn't this be dark:shadow? I'm not experienced with tailwind but the syntax looks weird to me

JustNireon (changes requested):
LGTM othere than the question i had about shadow. Tried to test as many notifications as i could think of but not sure if i had everything.

JustNireon (approved):
LGTM, Last commit removed unused class

PR #844, opened 2021-06-24
Remove fuzzily

Riscky:
This PR aims to replace fuzzily by postgres full text search

TODO: 
- [x] Remove Fuzzily from the dependencies
- [x] Remove all Fuzzily dependent code
- [x] Remove the trigrams table (migration)
- [x] Use the pg_search gem
- [x] Enable search for the members table
- [x] Enable database index
- [x] Enable trigrams (optional)
- [x] Also search for phone numbers, email addresses and student ids with fuzzy search
- [x] Remove the old trigrams rebuild script
- [x] Remove trigram search from infix (https://github.com/Casecommons/pg_search#limiting-fields-when-combining-features)

The workings of filters have NOT been touched and should work like they currently do (which is barely)

It is advisably to test this on more productionlike data before deploying, since it is hard to see the effects on search quality and speed for just 50 people.

Note for deploying: remove the systemd task that triggers the nightly trigram rebuild

PR #860, opened 2021-07-31
Debt/minor improvements

martijncasteel:
Some minor things. I tried to update the vulnerabilities but was unable to update the used ruby version with nix. I have trouble finding documentation, I was able to update the Gemfile locally but nix-shell kept using older ruby version.

### DIFF
--- a/app/models/member.rb
+++ b/app/models/member.rb
@@ -197,8 +197,8 @@ class Member < ApplicationRecord
 
   # Functions starting with self are functions on the model not an instance. For example we can now search for members by calling Member.search with a query
   def self.search(query)
-    # If query is blank, no need to filter. Default behaviour would be to return Member class, but we override by passing all active and studying members
-    return where(:id => (Education.select(:member_id).where('status = 0').map(&:member_id) + Tag.select(:member_id).where(:name => Tag.active_by_tag).map(&:member_id))) if query.blank?
+    # return all active members if query is empty
+    return Member.active if query.blank?
 
     # Otherwise we apply the filters and perform a fuzzy search on full name, phone number and email address
     records = filter(query)
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -53,7 +53,7 @@ class Activity < ApplicationRecord
     # Strip an activity name from all characters banks do not support
 
     # Remove all non-ascii characters (no language extension)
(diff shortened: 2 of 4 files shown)


app/models/member.rb
```diff
@@ -194,8 +194,8 @@ def self.search(query)
     phone_number = query.match(/^(?:\+\d{2}|00\d{2}|0)(\d{9})$/)
     return where("phone_number like ?", "%#{ phone_number[1] }") unless phone_number.nil?
 
-    # If query is blank, no need to filter. Default behaviour would be to return Member class, so we override by passing all
-    return where(:id => (Education.select(:member_id).where('status = 0').map(&:member_id) + Tag.select(:member_id).where(:name => Tag.active_by_tag).map(&:member_id))) if query.blank?
+    # return all active members if query is empty
+    return Member.active if query.blank?
```
Riscky, line 198: I think this is incorrect, the current implementation displays all studying AND active members.

Riscky (changes requested):
One comment, rest looks good :)

Riscky:
> I have trouble finding documentation, I was able to update the Gemfile locally but nix-shell kept using older ruby version.

Most Nix documentation can be found on nixos.org, specifically https://nixos.org/manual/nix/stable/
I wrote a small guide on dependency updagtes for Koala in [DEPENDENCIES.md](https://github.com/svsticky/constipated-koala/blob/development/DEPENDENCIES.md). 
Nix has a bit of a steep learning curve, I'm happy to help with any questions that come up.

Upgrading to the next major version of Ruby cannot be done before replacing Fuzzily (PR here: #844), since it is incompatible with Ruby 3.0.

app/models/member.rb
martijncasteel: Hi Rick, below you'll see the code `Member.active` is referring to, I believe it's the same line of code. 

https://github.com/svsticky/constipated-koala/blob/69ab067096798328bb6c9ea44cd830f6147264a5/app/models/member.rb#L74-L83

Do you mean by active currently in a committee? I don't think there is a query for that yet

Riscky: > Do you mean by active currently in a committee? I don't think there is a query for that yet

Yeah, I though that it meant that (and maybe this method should be renamed to avoid confusion in the feature)
You're right, your change is indeed correct.

Riscky (approved):

martijncasteel:
Thanks!

PR #861, opened 2021-08-01
Updated dependencies that are marked as vulnerable

Riscky:
This PR updates some important gems with vulnerabilities.\
Some migrations were added by `activestorage`.

~Since we update `activesupport` in this PR, we cannot perform the update before having replaced Fuzzily, which apparently relies on the now removed `normalize` method in `activesupport` (see: https://github.com/rails/rails/blob/v6.1.0/activesupport/CHANGELOG.md).~
This is done already :)

Other issues might pop up, please check this PR thoroughly.


JustNireon (commented):
LGTM Tried testing everything and found a few deprecated functions / issues.  One issue that should be resolved after this branch is merged is transitioning from autoLoad :classic to :zeitwerk
``
DEPRECATION WARNING: Initialization autoloaded the constants GhostscriptPreviewer, ApplicationHelper, Doorkeeper::DashboardHelper, DeviseHelper, FontAwesome5, FontAwesome5::Rails, FontAwesome5::Rails::IconHelper, ApplicationController, ApiController, RedirectUriValidator, and ImpressionistController.

Being able to do this is deprecated. Autoloading during initialization is going
to be an error condition in future versions of Rails.

Reloading does not reboot the application, and therefore code executed during
initialization does not run again. So, if you reload GhostscriptPreviewer, for example,
the expected changes won't be reflected in that stale Class object.


`config.autoloader` is set to `classic`. These autoloaded constants would have been unloaded if `config.autoloader` had been set to `:zeitwerk`.

In order to autoload safely at boot time, please wrap your code in a reloader
callback this way:

    Rails.application.reloader.to_prepare do
      # Autoload classes and modules needed at boot time here.
    end

That block runs when the application boots, and every time there is a reload.
For historical reasons, it may run twice, so it has to be idempotent.

Check the "Autoloading and Reloading Constants" guide to learn more about how
Rails autoloads and reloads.
 (called from instance_exec at /nix/store/9qqdb64pyijh2zrbpjznj843nnr9bbm3-ruby2.7.3-railties-6.1.4/lib/ruby/gems/2.7.0/gems/railties-6.1.4/lib/rails/initializable.rb:32)
``

Riscky:
New changes look alright, if you have tested them you can merge

JustNireon (approved):
LGTM, Had to rebase to resolve conflict in brakeman.ignore

PR #862, opened 2021-08-01
updated nixpkgs to 21.05 branch

Riscky:
This updates the pinned `nixpkgs` branch, which in turn leads to the following updates:
- nodejs is updated from version 12 to 14
- ruby is updated from version 2.6 to 2.7
- imagemagick is updated from version 6 to 7

No other major versions have changed.
This should be checked somewhat thoroughly, as there might be deprecations in above packages.

This change is also in some of my other PR's, but this one is probably easier to merge.


Riscky:
This is already in #844

PR #864, opened 2021-08-02
update rb-inotify to ruby 3.0 compatible version

### LINKED ISSUE #863: Deprecation warning: `rb_save_level`
In Ruby 2.7 some methods are deprecated which are removed in Ruby 3.0.
We need to solve the deprecation warnings before we can upgrade.
The warning is in the `rb-inotify` package:

```
rb-inotify/watcher.rb:66: warning: rb_safe_level will be removed in Ruby 3.0
```

Riscky:
This update seems to fix #863, although I cannot find the change back in the changelogs of `rb-inotify`


JustNireon (approved):
LGTM

PR #866, opened 2021-08-08
Feature/enable search transactions

JustNireon:
Initially when transactions got added i didn't add the feature to search for transactions since the old search engine was soon to be removed. Since this has happened this feature should be added asap.

Possible TO-DO's:

- [ ] Search on activity name (Pretty sure this is close to impossible since we only store activity id's in a list)
- [ ] Add date range as possible search parameter
- [ ] When pressing enter you should search and not open the payment method list


Riscky:
> Add date range as possible search parameter

I think that this should be a filter instead of a search param

Riscky (approved):
very nice, works beautifully :)

PR #868, opened 2021-08-11
update update function

JustNireon:
When merging #861 I created a new bug that made it ignore some validations


Mstiekema (approved):
lgtm

PR #873, opened 2021-09-01
Fix task status:mail query

### LINKED ISSUE #872: task status fails
The current rake task 'status' fails due to an invalid SQL query. This is probably due to the migration to postgresql and should be a trivial fix.
/lib/tasks/status.rake to look at the invalid task

JustNireon:
Resolves #872


Riscky:
LGTM

Riscky (approved):

PR #876, opened 2021-09-10
Speed up production and nix-shell

Riscky:
A whole lot of thought went into this PR, and this resulted in the following:

- I reverted the assets precompilation in Nix. This slowed down nix-shell considerably since it is not cached. This can be in Nix again when we go full nix build
- When run in production, the precompilation phase links the compiled js and css to default locations.
- When run in production, webpacker is bypassed and the default locations are used. This delivers a considerable speedup, since webpacker is just ignored completely

Also I included something I just came across
- The mysql2 gem is removed, which should not be in use anywhere anymore

I also created an issue to move to serving gzipped js and css in the future, which should give even more speedup: #878
This PR does not need any changes in sadserver to be deployed succesfully :)

Putting asset precompilation back into nix is on my planning, but that is probably more easy with nix-run (I got nix-run working in another branch, so maybe that will land sometime soon)

- [x] I should probably add some long comment on the right places to explain what I've done


JustNireon (approved):
LGTM

PR #880, opened 2021-09-15
FIX: Pin docker compose version to 13

J00LZZ:
The postgres version in the docker compose file was unpinned, while this is not an issue right now, it will be when docker sees 14 as the latest.


Riscky:
:shipit:

Riscky (approved):

PR #881, opened 2021-09-15
Schema format for application.rb

JustNireon:
schema_format was only set for development. This should be set for all environments


Phavion (approved):
LGTM!

PR #852, opened 2021-07-06
Enhancement/payconiq payment activity listing description

Raymonzut:
Fixes #831

- Continuation of #851

PR #883, opened 2021-09-17
Payconiq payment activity listing description implementation: Method call, join_with_char_limit, should be a static call

Raymonzut:
Hotfix on #852


JustNireon (approved):
LGTM

PR #897, opened 2021-10-26
Add tailwind jit purge compilation (and update to webpacker 6)

jasperro:
JIT mode is still in preview but speeds up build.
This is not compatible with semantic versioning though.
CSS bundle size is reduced 20x as purge was not working correctly before.


Riscky:
Nice, I won't mind getting this merged

> This is not compatible with semantic versioning though.

I'm not sure what is meant here, can you explain how this breaks semantic versioning?

> purge was not working correctly before.

I think it is, bundles in production are (at least last time I checked) quite small. We disabled purging in development because it was too slow, which your patch fixes.

We prebuild the bundles in production, so I think this should also work there

jasperro:
> I'm not sure what is meant here, can you explain how this breaks semantic versioning?

As the JIT version is still in development, there could be small chance breaking changes.

The tests also need to be fixed, where would this need to be done? It seems unclear to me why the tests are failing.

jasperro:
> I'm not sure what is meant here, can you explain how this breaks semantic versioning?

As the JIT version is still in development, there could be small chance breaking changes.

The tests also need to be fixed, where would this need to be done? It seems unclear to me why the tests are failing.

Riscky:
> The tests also need to be fixed, where would this need to be done? It seems unclear to me why the tests are failing.

The tests seem to have been cancelled, maybe they were taking too long or Github did an oopsie.
I have tried to rerun the tests, but you will have to rebase your branch on the latest development before the tests start working again because of svsticky/sadserver/issues/333 (fixed in #902, which is not in your branch yet)

Riscky:
> As the JIT version is still in development, there could be small chance breaking changes.

This is fine btw, we pin everything with Nix (except that we forgot to pin Nix itself in our CI, ironically) so updates are always a manual action anyway

Riscky:
> Running dotenv rails assets:precompile --trace hangs, seems like webpacker never actually finishes in production/testing. It only works with the rails s dev server

That sounds like it could be a problem with your changes, since webpack(er) also packs tailwind

btw we are changing our branching structure (see svsticky/sadserver/issues/323) so you will have to change this PR to merge to master instead of development :)

postcss.config.js
```diff
@@ -1,23 +1,13 @@
...
+      stage: 3,
     }),
-    // don't purge css in development
-    process.env.NODE_ENV === 'development' ? null :
-    require("@fullhuman/postcss-purgecss")({
-      content: [
-        "./app/**/*.haml",
-        "./app/javascript/**/*.js"
-      ],
-      defaultExtractor: content => content.match(/[^<>"{\.'`\s]*[^<>"{\.'`\s:]/g) || []
```
Riscky, line 20: I think this extractor was written to deal with the syntax of haml files (I think this was the source: https://medium.com/expected-behavior/purgecss-extractors-for-haml-a9dfe23a3504)
Can you verify that the correct classes are still extracted?

Riscky:
> Running dotenv rails assets:precompile --trace hangs, seems like webpacker never actually finishes in production/testing. It only works with the rails s dev server.

Riscky:
me

jasperro:
I think it is best to wait with this commit until webpacker 6 (with postcss 8) releases. The last commit I pushed was incompatible. It is possible that the postcss 7 compatibility layers has some issues with jit purge.

postcss.config.js
jasperro: In the most recent commit I added the same default extractor to the tailwind config, it should all work correctly. webpack-dev-server quickly reloads on change, which is reflected in the css.

jasperro:
I think this is ready to be merged, I don't think I really check the devise? login screen, but doorkeeper works correctly.

This PR also includes a update to webpacker 6 with webpack 5.

jasperro:
On this branch I'm testing it out on the members pages too: [jasperro/constipated-koala:feature/redesign-main](https://github.com/jasperro/constipated-koala/tree/feature/redesign-main). Seems to be working pretty well, with all the styles getting picked up.

app/javascript/application.js
```diff
@@ -19,11 +19,11 @@ import "bootstrap";
 require("@rails/ujs").start();
```
Riscky, line 19: Merging all packs into one means we are shipping more javascript then strictly necessary for different parts of the application.
I see why we would want this change (it is easier to manage, packs are cached across applications parts), but can you verify how much bigger the pack gets?

app/views/admin/members/payment_whatsapp.html.haml
```diff
@@ -0,0 +1,7 @@
+= I18n.t('admin.payment.whatsapp.title', user: current_user.credentials.first_name)
```
Riscky, line 1: Very nice to see that we can do away with `erb-loader`, that is going to save us some trouble going forward.
Can you also remove the package itself?

config/webpacker.yml
```diff
@@ -86,28 +89,11 @@ test:
   # Compile test packs to a separate directory
   public_output_path: packs-test
 
-
-staging:
```
Riscky, line 90: Does the removal of the `staging` section mean that the production config is automatically picked on the staging server?
Or does that need a configuration change on the staging server (setting `NODE_ENV=production` for example)?
(I'm in favor of removing the staging settings btw, I like our staging environment to be as close as possible to production)

postcss.config.js
Riscky: thanks :)

Riscky (approved):
I'm really happy with this PR: less moving parts, updated packages and a better developer experience!
I have a few small questions, but feel free to merge this :+1:

Riscky:
> On this branch I'm testing it out on the members pages too: jasperro/constipated-koala:feature/redesign-main.

That sounds amazing! I have some ideas for redesigning some parts, if you are interested I'll shoot in some issues to show it.

app/views/admin/members/payment_whatsapp.html.haml
jasperro: erb-loader is still being used for I18n-js, not it can't be removed. It generates a js file with the template that includes the translations.

config/webpacker.yml
jasperro: I'm not sure, I just adapted the default webpacker 6 config. Guess I missed the change. I think that the node env (maybe rails env too?) may have to be changed on the server from staging to production to make it work, but I don't know for sure.

app/javascript/application.js
jasperro: The default directory for webpacker (src) packs was changed from app/javascript/packs to app/javascript. The PR reflects this. I don't think it changes the way the packs get sent to the clients though. Webpacker still seems to give multiple output js files.

Riscky: Ah you're right, my bad

config/webpacker.yml
Riscky: We'll see when we deploy 🤷

PR #902, opened 2021-11-03
Pin nix to version 2.3 to keep builds from failing

Riscky:
Due to the release of Nix 2.4 (see svsticky/sadserver/issues/333), and the fact that the install-nix-action just updates Nix without warning, we need to pin the version of Nix in CI for the time being.


Phavion (approved):
LGTM!

PR #907, opened 2021-11-10
Payconiq removal

### LINKED ISSUE #900: Payconiq is stopping, remove payconiq from the codebase
This is a hard pill to swallow (Especially for @JustNireon, who put so much time in the implementation), but Payconiq is stopping its services per January 1, 2022.
We will have to remove the Payconiq code from the codebase, without breaking the payment logs (which we will have to store for 7 year according to law).
This should not be too difficult, and maybe we can split up the change, first removing the client side code and clean up the backend stuff later.

We might have to discuss when and how to roll out this change, suggestions are welcome.

A whole other discussion is to find a suitable replacement for payments in the magnificent-sloth app, but this will have to go through @Jelte-Akker first (and also this is not the right repository for discussing this).

Again I want to thank @JustNireon for his diligent work on implementing Payconiq support, I think Koala is (even after removing it) in better shape; iDeal payments are now possible, the interface has been improved, and the architecture is capable of adding new payment methods.

JustNireon:
Resolves #900,

This removes all references to Payconiq in the codebase (where possible)
This also removes the dropdowns that once showed the choice between Ideal and Payconiq, but are now useless.
This should be tested thoroughly on staging, as test payments cannot go through in development (due to the return address not being reachable by the payment provider).

### DIFF
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -138,9 +138,11 @@ class Payment < ApplicationRecord
   def transaction_fee
     case payment_type.to_sym
     when :ideal
-      return Settings.mongoose_ideal_costs
+      Settings.mongoose_ideal_costs
+    when :payconiq_online
+      0
     when :pin
-      return 0
+      0
     end
   end
 
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -12,7 +12,8 @@ class Payment < ApplicationRecord
   validates :payment_type, presence: true
 
   enum status: { failed: 0, in_progress: 1, successful: 2 }
-  enum payment_type: { ideal: 0, pin: 3 }
+  # Keep payconiq_online because it is still present in the database
+  enum payment_type: { ideal: 0, payconiq_online: 1, pin: 3 }
   enum transaction_type: { checkout: 0, activity: 1 }
   belongs_to :member
 
--- a/app/models/payment.rb
+++ b/app/models/payment.rb
@@ -4,7 +4,7 @@ class Payment < ApplicationRecord
 
   self.primary_key = :token
 
-  attr_accessor :issuer, :payment_uri, :message, :payconiq_qrurl, :payconiq_deeplink
+  attr_accessor :issuer, :payment_uri, :message
 
   validates :description, presence: true
   validates :amount, presence: true, numericality: true
@@ -12,7 +12,7 @@ class Payment < ApplicationRecord
   validates :payment_type, presence: true
 
   enum status: { failed: 0, in_progress: 1, successful: 2 }
-  enum payment_type: { ideal: 0, payconiq_online: 1, payconiq_display: 2, pin: 3 }
+  enum payment_type: { ideal: 0, pin: 3 }
   enum transaction_type: { checkout: 0, activity: 1 }
   belongs_to :member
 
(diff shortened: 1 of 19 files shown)


Riscky:
I checked, and I'm pretty sure we don't have to change anything in the database schema.
IMHO we can put this on staging and start testing

Riscky:
I rebased on the latest master, fixed the CI failures (linting, brakeman), removed some more implicit references to  Payconiq and updated the OP.
I think the PR should be ready for testing now

app/models/payment.rb
```diff
@@ -4,23 +4,23 @@ class Payment < ApplicationRecord
...
+  attr_accessor :issuer, :payment_uri, :message
 
   validates :description, presence: true
   validates :amount, presence: true, numericality: true
   validates :status, presence: true
   validates :payment_type, presence: true
 
   enum status: { failed: 0, in_progress: 1, successful: 2 }
-  enum payment_type: { ideal: 0, payconiq_online: 1, payconiq_display: 2, pin: 3 }
+  enum payment_type: { ideal: 0, pin: 3 }
```
Raymonzut: Good to see the variants still retaining their original numbers for backwards-compatibility,  but be careful

db/seeds/payments.rb
```diff
@@ -7,7 +7,7 @@
 
   15.times do
     transactiontype = Faker::Number.within(range: 0..1)
-    paymenttype = Faker::Number.within(range:0..3)
+    paymenttype = Faker::Number.within(range:0..1)
```
Raymonzut: , as paymenttype variants are 0 and 3, not 0 and 1 ;D

db/seeds/payments.rb
```diff
@@ -7,7 +7,8 @@
 
   15.times do
     transactiontype = Faker::Number.within(range: 0..1)
-    paymenttype = Faker::Number.within(range:0..3)
+    # either ideal or pin (see app/models/payment.rb)
+    paymenttype = Faker::Number.within(range: 0.0..1.0) < 0.5 ? 0 : 3
```
Raymonzut, line 11: which you did correctly here :+1:

app/models/payment.rb
```diff
@@ -190,10 +139,7 @@ def transaction_fee
     case payment_type.to_sym
     when :ideal
       return Settings.mongoose_ideal_costs
-    when :payconiq_online
-      return Settings.payconiq_transaction_costs
-    else
-      # :pin, :payconiq_display
+    when :pin
```
Raymonzut, line 142: Shouldn't it also be `0` on `:payconiq_online` till it is removed, as it could still be present in the database?

Raymonzut (changes requested):
Generally looks good, I propose the following to discuss:
- [ ] Seeds: Reconsider the Faker generators for `paymenttype`
- [x] Transaction fees: will `:payconiq_online` still need to be accounted for in the fees, as they could be in the old DB?

app/models/payment.rb
Riscky: good point, I'll add that

Riscky:
I have updated the fees calculation. The seeds generator is already correct I think, your comment is on an outdated version of the code.

Raymonzut (dismissed):
LGTM, all pointers from last review were addressed and added :D

- seeds => has been fixed since, now outdated
- fees calculation: corrected

JustNireon:
Thanks for the help! I've been very busy so the fixes are appreciated

Riscky:
Thanks for the review!

PR #908, opened 2021-11-16
Removed the hack that circumvented webpack in production

Riscky:
In a past PR (#876) I used a trick to circumvent webpacker (and webpack) in production environments.
With the latest version of webpack (see #897) this trick has become a lot more difficult, so I took it out of commission.
This PR just reverts the changes, which should fix the javascript in production environments.


Phavion (approved):
Big sad dat t weer breekt :(, verder LGTM

PR #917, opened 2021-12-02
Whatsapp endpoint broken

JustNireon:
Not sure why everything was changed from erb to haml but this was broken.


Riscky (changes requested):
Can you import it as Haml instead?

Riscky (approved):
joe

Riscky (approved):
:shipit:

Issue #386, opened 2018-10-03
Webhooks?


maartenberg:
I've half-seriously talked about this in the meetings before, but I seriously think that adding outgoing webhooks for some events in Koala could be a good idea to allow providing functionality without adding service-specific API links to Koala. Some examples:

- If a member's email changes, the email would be (ideally) also updated in the Mailchimp lists, or a board member should get an email with the previous and new address, to get a reminder that the email address needs to be updated.
- If an activity is created or changed, a signal could be sent to Google Calendar to create or update a matching activity.

If we had webhooks, this could then be done without coupling Koala itself to an external service we might stop using, and we could instead write some separate script that translates Koala's data to the API of whatever service that needs to be updated.

This is probably a big idea etc., so yeah. Thoughts?

Riscky:
> usecase for webhook is unclear

We found a usecase for undead mongoose :)

Riscky:
Closed by #921

PR #921, opened 2021-12-09
API stuff for undead-mongoose

Phavion:
This PR will contain all changes needed to get a functioning undead-mongoose.

- [x] Create route to obtain user information for registering.
- [x] Create simple webhook to trigger undead-mongoose to update a user.
- [x] Do we want a temporary way in which undead-mongoose transfers 'tegoed'?


Riscky:
You can close #386 with this PR :)

Riscky:
We should also (maybe not in the PR, but do make issues if that is decided), communicate with Checkout when a member is deactivated or deleted.
Deactivation here means either end-of-study or suspension.

Mstiekema (approved):
lgtm

PR #927, opened 2022-01-02
Checkmark consistency and overrides for enrollable date

Plopmenz:
The "is_enrollable" checkmark now gets automatically checked when an activity opens. This checkmark can then be unchecked to close the activity again. Checking the "is_enrollable" checkmark before the activity is opened, will remove the open date and time and open up the activity.

Creating an activity with both an open date and time and the "is_enrollable" checkmark will also remove the open date and time and open the activity for enrollement. Perhaps an error should be given to warn the creator instead?

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -193,18 +193,18 @@ class Activity < ApplicationRecord
   end
 
   def open?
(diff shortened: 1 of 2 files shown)


Riscky (approved):
Nice work!

PR #928, opened 2022-01-03
Payconiq removal hotfix

JustNireon:
Due to having time constraints i've been unable to fully remove payconiq before theirs system stopped. This fix will remove the possibilty of starting a payconiq transaction which is the only breaking part right now. Somebody else can pick up #907 or ill continue it when i have time again


Riscky (approved):
LGTM, do you have time to deploy?

PR #931, opened 2022-01-18
Update tailwindcss to version 2 to 3

jasperro:
Now with JIT as default mode, and the purge/content config doesn't need the custom defaultExtractor anymore to work.


Riscky (approved):
nice work!

PR #935, opened 2022-01-28
Update packages to Ruby 3

Riscky:
This updates the packages to Ruby 3
Some packages needed updating to support this, which is reflected in the gemfile changes


jasperro:
I'm currently having some issues with i18n-tasks and ruby_dep not being compatible with ruby 3 (when running bin/webpack-dev-server). i18n-tasks just needs to be upgraded to 0.9.37. ruby_dep seems to be a dependency of listen. listen needs to be upgraded to `"gem 'listen', '~> 3.1', '>= 3.1.5'"`
Also having issue with `bin/webpack-dev-server`:
It shows error:
```
Could not find rails-6.1.4 in any of the sources
Run `bundle install` to install missing gems.
```
Not sure if this has something to do with the ruby version.

Downgrading ruby in default.nix to v2 is working as a temporary fix for both issues.

Riscky:
I have updated the `i18n-tasks` and `listen` gems.
Apparently `./bin/webpack-dev-server` does not work on master either, so I'll try to get a fix out for that issue as soon as possible.
In the meantime `bundle exec bin/webpack-dev-server` should work.

PR #944, opened 2022-02-10
Fix #939

### LINKED ISSUE #939: Force email change is broken
Reproducing the issue:
- Change the email address of test@svsticky.nl (or any member with an account, but in development this is the only member with an account)
- The page now shows "A confirmation mail is send to test2@svsticky.nl, once confirmed the e-mailadress will be updated. Or [force](https://koala.dev.svsticky.nl/members/3/force_email_change) the change here."
- Click force, and see the following error: " ActionController::ParameterMissing in Admin::MembersController#force_email_change "

Riscky:
There are no params in a patch, we should check member attributes instead.
Also reversed the order so this is more testable (Mailchimp is not available in staging / CI)


Riscky:
We should in the future move the mailchimp update code to the 'after_save' method (see #921 for an example), so we have the mailchimp code in one place.

Raymonzut (commented):
LGTM

PR #946, opened 2022-02-21
add deleted_at to users

JustNireon:
add deleted_at column to users this makes it possible to soft delete admin users so they cannot login anymore but the accounts are still there for logging


Riscky:
LGTM

PR #952, opened 2022-03-26
Rewrite email dropdown code

### LINKED ISSUE #942: Sentry: NoMethodError permit! in Admin::ParticipantsController
`undefined method `permit!' for nil:NilClass`

in `app/controllers/admin/participants_controller.rb` in `mail` at line `52`

Probably happens when trying to send an email to an empty list of participants

Riscky:
This should fix #942 by disabling the send button when no recipients are selected.
The section has been completely rewritten to be more readable.

### DIFF
--- a/app/javascript/src/lib/mail.js
+++ b/app/javascript/src/lib/mail.js
@@ -94,27 +94,28 @@ import toastr from "toastr";
       $(this)
         .find("#recipients select")
         .on("change", function (event, selector) {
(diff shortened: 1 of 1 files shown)


Raymonzut (dismissed):
LGTM
+ saves duplication
+ (j)query on `this` indeed gives the selector you want in this context anyways, only as a fallback foreach-case is redundant

Issue #940, opened 2022-02-08
Sentry: ArgumentError: comparison of ActiveSupport::TimeWithZone with nil failed


Riscky:
`app/models/activity.rb` in `<` at line `219`

Probably solved with some input sanitation on the activity edit page?

PR #953, opened 2022-03-26
activity edit: check if start_time is present before reading its value

Riscky:
This should fix issue #940 by checking if the start_time is present before reading its value.
While debugging the issue I came across a related issue that also has been fixed.


Raymonzut (approved):
LGTM,

ez fix for the conditional err & conditional view

PR #961, opened 2022-04-06
fix missing translation for payconiq_online

Riscky:
This translations were removed in #907, but are still shown on the transactions page.
This is still needed as long as there are Payconiq transactions in the database (which is required by Dutch law for 7 years)


Raymonzut (approved):
LGTM, mentioned in #907 we want to keep `:payconiq_online`, so it follows that the translations should be kept

PR #962, opened 2022-04-06
Show correct time on posts

### LINKED ISSUE #959: Board posts: published at time is incorrect
### Expected behaviour
The correct time is shown

### Actual behavior
The hours are shown correctly, but the minutes are always set to 04, no matter the actual published time

### Steps to reproduce
- both admin and user

Riscky:
This fixes #959 by setting the correct format string


Raymonzut (approved):
LGTM

PR #965, opened 2022-04-14
Fix incorrectly formatted time on board posts

### LINKED ISSUE #959: Board posts: published at time is incorrect

Riscky:
Fixes #959
I though I fixed this already in #962, but apparently the bug was also present at the members side.


Phavion (approved):
LGTM

PR #966, opened 2022-04-14
Fix internal controller after big rubocop oof (fix rubocop too)

Phavion:
Fixes internal api controller and removes erroneous check from rubocop


Riscky:
LGTM

Riscky (approved):

PR #967, opened 2022-04-14
prevent the CI from running twice on a PR

Riscky:
Currently the CI is instructed to run both on a push and a pull request, which leads to double runs.
This PR prevents that.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,5 +1,6 @@
 name: "Build & Test"
-on: [push, pull_request]
+on:
+  - push
 jobs:
   build:
     runs-on: ubuntu-latest


Raymonzut (approved):
LGTM, fits our rebase workflow

Raymonzut:
So apparently it still does it once too many sadly, so here are some ideas:

- Define `branches-ignore`'s, for master perhaps
- Still do `push` and `pull_request`, but only on certain types of them (like `created`)

@Riscky what do you think?

Riscky:
I don't think it's a big problem to run the tests again after the PR is merged, at least not as important as the current fix.
Ignoring master would also break the badges in the readme (reminds me that I need to remove the development branch badge).

Raymonzut:
Great, then we will leave it at this reduction :D

PR #969, opened 2022-04-15
make it possible to unenroll after the deadline for reservists

### LINKED ISSUE #964: Un-enrolling after cancel date when on the spare list not possible
### Expected behaviour
Can undo enrolling after cancel date when on the spare list

### Actual behavior
Not allowed, treated the same as enrolled in the activity (not on the spare list)

### Steps to reproduce
- admin/user
- browser: any
- enroll on full activity, on spare list
- try to opt-out

Riscky:
Fixes #964 

This makes it possible for a reservist to unroll from the spares list, even after the unenroll deadline has passed.
This lets the member signal that they don't want to come to this activity, even if extra spaces are available.


Raymonzut (approved):
LGTM, tested and works as expected, good that you also added the check again at the confirm step :D

PR #971, opened 2022-04-15
limit export of translations with i18n to actually used ones

### LINKED ISSUE #955: I18n-js export more translations than neccessary
In the i18n-js export (generated at `app/javascript/src/translations.js`) we find all sorts of translations that are not actually used in the frontend (For example an exhaustive list of countries for some reason).
Webpack is not smart enough to filter this data, so we actually export the whole bunch to each pack we build.
We should not do this.

There is an  i18n-js settings file at `config/i18n-js.yml`, in which we can add filters.
Explanation for the filters can be found at https://github.com/fnando/i18n-js (note that we use version 3.9.1, not the alpha version of 4.0).

Riscky:
fixes #955 

This instructs `i18n-js` to limit the exported translations to the keys we actually use in javascript.
This doesn't prevent exporting some translations we don't use, but at least limits the size of the file with a factor 50 (1.6M to 34K)

### DIFF
--- a/config/i18n-js.yml
+++ b/config/i18n-js.yml
@@ -11,5 +11,7 @@ translations:
     prefix: "import I18n from 'i18n-js';\n"
     suffix: "export default I18n;\n"
     only:
-      - "en"
-      - "nl"
+      - "*.admin"
+      - "*.checkout"
+      - "*.mailings"
+      - "*.members"


Raymonzut (approved):
LGTM, also easy to extend translation domains as needed

PR #977, opened 2022-04-19
make the activities email send page work with the new editor

### LINKED ISSUE #974: Editor for emails is broken
On the admin activities pages an email editor can be found.
This includes a bar with text option (for example to make a bit of text bold).
This editor bar is broken.
If this is not an easy fix, we might as well use Quill here immediately (see #886)

Riscky:
fixes #974

This fixes the editor loading code and html, and seems to work alright.
I haven't checked yet how the email look like


Mstiekema (approved):
LGTM!

PR #981, opened 2022-04-21
create payment request mail task

Riscky:
This PR adds a new mailing, with a rake task that should be run on a schedule (see svsticky/sadserver/pull/390 for the accompanying server PR).
This email summarizes all open payments and asks users to pay this asap via the payment portal.
The email looks like this:
image

Unlike the other mailings, the text version of the email is rendered from a template instead of a heredoc, which in my opinion is cleaner.

This PR also changes some things outside of the mailers:
- The unused `Member.payable_unpaid_activities` has been removed
- `Member.debtors` has been added, which returns a list of all members that have open payments.
- `Member.participant_by_activity` has been added, which returns the participant object that belongs to the member and the given activity. This method should be used carefully: don't pass an activity that the member is not a participant of.
- If the member does not have a user account (and thus no language preference set), the language preference is defaulted to Dutch.
- The check if a Mailgun token is available in the environment has been moved to the common `mail` and `mails` methods.

This PR does not:
- Change the definition of defaulters (wanbetalers) to people to that ignored the new payment emails for a month. The current definition of defaulters is something like: people that received 4 'inform' emails.
- Change the default text for the 'inform' emails to something unrelated to payments.
- Remove the dropdown option for 'no_payers' ('NOG BETALERS') in the 'inform' email editor.
- Update the definition of unpaid activities on the admins' payment page to align with the definition used in the email (`debtors`).
- Add a list of _expired, but not yet set to payable_ activities to the admins' payment page (which helps the treasurer to find those activities).
- Move all mailings to using text templates.

All of these should probably be done sometime in the future, but seem out of scope for the current PR (TODO: make issues for these things).


config/locales/mailings.en.yml
```diff
@@ -63,6 +63,11 @@ en:
         activity_payable_unvisited_html: You are going to visit <strong>%{subject}</strong>, this activity costs %recipient.price%.
         activity_visited_html: You visited <strong>%{subject}</strong>, this activity cost %recipient.price%.
         payment_instructions_html: Please pay this as soon as possible via the <a href="https://koala.svsticky.nl/member/payments">payment portal</a>
+    payment:
+      open_activities: 'You have the following unpaid activities:'
+      payment_instruction: Pleas pay this as soon as possible via the <a href="https://koala.svsticky.nl/member/payments">payment portal</a>.
```
Raymonzut, line 68: Typo, duplicate translation with the ones above, maybe join them together as one?

app/views/mailings/payment_mailer/requestmail.html.haml
```diff
@@ -0,0 +1,25 @@
...
+  = I18n.t('mailings.payment.total_of', amount: total)
+
+%p
+  = raw I18n.t('mailings.payment.payment_instruction')
+
+%p
+  = I18n.t('mailings.best_regards')
+
+%p
+  = I18n.t('mailings.signature')
```
Jelte-Akker, line 25: Please make sure that the the mail is from the treasurer and that replies are directed towards penningmeester@svsticky.nl

Jelte-Akker (dismissed):
lgtm :)

Jelte-Akker (approved):
lgtm

PR #982, opened 2022-04-21
fix a few easy haml lints

Riscky:
This enables and fixes some easy to solve haml-lints.
There are a lot of more useful lints that are not yet enabled, but lets start with these.


Raymonzut (approved):
LGTM, the ignore of the `ConsecutiveComments` rule also fixes CI for #981 ;)

PR #984, opened 2022-04-21
fix some rubocop-rails todo's

Riscky:
This fixes some rubocop-rails todo's.
The most important of those is that we enable the `DynamicFindBy` cop.
This cop disallows the use of the confusing dynamic `find_by_x` method, in favor for the clearer `find_by x:`.


Raymonzut (approved):
LGTM

PR #988, opened 2022-04-25
Remove pin payment option from intro.svsticky.nl payment dropdown

### LINKED ISSUE #986: Remove pin payment option from intro.svsticky.nl

Raymonzut:
Fixes #986

### DIFF
--- a/app/controllers/public/home_controller.rb
+++ b/app/controllers/public/home_controller.rb
@@ -21,8 +21,8 @@ class Public::HomeController < PublicController
     total = 0
 
     # if bank is empty report and test model for additional errors
-    flash[:error] = nil
-    flash[:error] = I18n.t(:no_bank_provided, scope: 'activerecord.errors.subscribe') if params[:bank].blank? && params[:method] == 'IDEAL' && @member.educations.none? { |education| Study.find(education.study_id).masters }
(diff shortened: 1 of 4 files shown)


Riscky:
Can you also update the backend code in `app/controllers/public/home_controller.rb`?
Otherwise it would still be possible to forge a request to get a 'pin/cash' membership.

app/controllers/public/home_controller.rb
```diff
@@ -21,8 +21,8 @@ def create
     total = 0
 
     # if bank is empty report and test model for additional errors
-    flash[:error] = nil
-    flash[:error] = I18n.t(:no_bank_provided, scope: 'activerecord.errors.subscribe') if params[:bank].blank? && params[:method] == 'IDEAL' && @member.educations.none? { |education| Study.find(education.study_id).masters }
+    flash[:error] = I18n.t(:deprecated, scope: 'activerecord.attributes.payments.payment-type') unless params[:method] == 'IDEAL' # Cash/PIN
```
Riscky, line 24: We can also completely remove the `method` form parameter, as it only has one option anyhow.
I don't think we are adding any other payment methods in the near future.

Riscky (dismissed):
Looks good, but I suggest we completely remove the `method` parameter.

PR #990, opened 2022-04-27
admin/member creation: remove superfluous and undefined NIL parameter

### LINKED ISSUE #978: Adding a new user results in 500 error
When you try to add a user on your own from the admin UI, you receive a 500. 

image

Riscky:
Fixes #978 

`NIL` is undefined, which is why the crash happens (the correct spelling is `nil`).
How this has been undiscovered for 2 years is a bit of a mystery to me.
The extra argument to `options_for_select` isn't needed, so I removed it.


Raymonzut (approved):
LGTM, quite hilarious it was present for so long :)

PR #998, opened 2022-05-09
updated participant filter text

### LINKED ISSUE #994: First/Second/Third/Master filter gives wrong warning to user
### Expected behaviour
Koala warns after enrolling for a yearfiltered activity that what group it is filtered for

### Actual behavior
Koala gives the wrong warning, it is always warning that it is a 'eerstejaars' 'freshmen' activity. 
When the user is the correct group, it does not give the warning, which is as it should be. 

### Steps to reproduce
- admin/user
- admin: make an activity that is filter for a yeargroup
- user: enroll for an activity outside your year
- browser:
- koala.dev.svsticky.nl

Riscky:
fixes #994

uses this magic: https://apidock.com/rails/Array/to_sentence
to append arrays together.

### DIFF
--- a/app/controllers/members/participants_controller.rb
+++ b/app/controllers/members/participants_controller.rb
@@ -90,18 +90,15 @@ class Members::ParticipantsController < ApplicationController
                                         activity: @activity.name)
     elsif !participant_filter_check?
       reservist = true
-      if !participant_freshman_check?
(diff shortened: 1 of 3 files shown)


Jelte-Akker (approved):
lgtm!

Jelte-Akker (approved):
top

PR #1001, opened 2022-05-11
Rubocop: Enforce the use of function parentheses, and auto-fix all offenses

Riscky:
This is an old pet peeve of mine: Ruby allows functions to be called without parentheses.
In my opinion, this just makes the code less readable, especially for beginners.
All changes are made my `rubocop --auto-correct`, which should be completely safe.
The exception to this rule is a somewhat unrelated change in  `app/controllers/public/status_controller.rb`, where 'not include' was replaced by 'exclude' by `rubocop --auto-correct-all`.
I double-checked, and this should be safe too.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -33,10 +33,10 @@ Metrics/MethodLength:
 
 # Allow 'some.variable == 0' and others
 Style/NumericPredicate:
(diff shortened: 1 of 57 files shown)


Raymonzut (approved):
LGTM

PR #869, opened 2021-08-17
Add attributes to activity for more accesibility

JustNireon:
Adds the functionality to limit enrollment to penultimate year and senior year for bachelors.
Additionally this also makes it possible to limit enrollment to combinations of the 4 filters (freshman,sophomore,senior,master). For example you could allow only freshman and masters to an activity.

fixes #418

PR #1002, opened 2022-05-14
allow non-sparelist enrolls to filterless activities

Riscky:
This solves a bug introduced in #869
It was not possible to enroll without being placed on the spare list for activities without filters.
This has been fixed in this PR.


Jelte-Akker (approved):
lgtm, works on staging!

PR #1005, opened 2022-05-20
Update the definition of defaulters to align with the payable attribute

Riscky:
This makes the definition of defaulters use the payable attribute.
This code is a lot more readable too.

### DIFF
--- a/app/controllers/admin/home_controller.rb
+++ b/app/controllers/admin/home_controller.rb
@@ -13,8 +13,6 @@ class Admin::HomeController < ApplicationController
 
     @unpayed = Participant.where(reservist: false, paid: false).joins(:activity).where('activities.start_date < NOW()').sum(:price) + Participant.where(reservist: false, paid: false, price: nil).joins(:activity).where('activities.start_date < NOW() AND activities.price IS NOT NULL').sum('activities.price')
 
-    @defaulters = Participant.where(reservist: false, paid: false).joins(:activity, :member).where('activities.start_date < NOW()').group(:member_id).sum(:price).merge( \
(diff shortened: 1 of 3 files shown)


app/controllers/admin/home_controller.rb
```diff
@@ -13,8 +13,6 @@ def index
 
     @unpayed = Participant.where(reservist: false, paid: false).joins(:activity).where('activities.start_date < NOW()').sum(:price) + Participant.where(reservist: false, paid: false, price: nil).joins(:activity).where('activities.start_date < NOW() AND activities.price IS NOT NULL').sum('activities.price')
 
-    @defaulters = Participant.where(reservist: false, paid: false).joins(:activity, :member).where('activities.start_date < NOW()').group(:member_id).sum(:price).merge( \
-      Participant.where(reservist: false, paid: false, price: nil).joins(:activity, :member).where('activities.start_date < NOW()').group(:member_id).sum('activities.price ')
-    ) { |_, sum_a, sum_b| sum_a + sum_b }.sort_by { |_, sum| -sum }.map { |k, v| [Member.where(id: k).select(:id, :first_name, :infix, :last_name).first, v] }.take(12).delete_if { |_, v| v == 0 }
+    @defaulters = Member.debtors.sort_by { |debtor| -debtor.total_outstanding_payments }.take(12)
```
Riscky, line 16: the minus does reverse sort, I also could use reverse() here, which is more readable, but a little slower.

Raymonzut (approved):
LGTM, better to reuse that instead of joining it on the fly

PR #1007, opened 2022-05-25
Add payable updated at

### LINKED ISSUE #999: Add a payment deadline per activity
When setting an activity to payable there should be an attribute like "payable_from" which is the moment the activity got payable. When sending the weekly emails behind every activity this date should be compared to a 28(?) deadline and see how many days are left. This way we get a good definition for debtors and give a better indication when activities should be payed

JustNireon:
closes #999


app/views/admin/payments/index.html.haml
```diff
@@ -52,17 +52,22 @@
                   = I18n.t('admin.activities.single')
                 %th
                   = I18n.t('admin.payment.days')
+                %th
+                  = I18n.t('admin.payment.times')
               %tbody
               - @last_impressions.each do |tupel|
```
Riscky, line 58: can you rename this to `tuple`?

app/controllers/admin/payments_controller.rb
```diff
@@ -1,25 +1,19 @@
...
-      ).last
-
-      days = if impression
-               Integer(Date.today - impression.created_at.to_date)
-             else
-               "-"
-             end
-      [activity, days]
+      if activity.is_payable_updated_at.to_date > Date.today.prev_occurring(:friday)
+        days = '-'
```
Riscky, line 9: I think string formatting should live in the view, not in the controller.

app/controllers/admin/payments_controller.rb
```diff
@@ -33,8 +27,7 @@ def index
 
     # Get members of which the activities have been mailed 4 times, but haven't paid yet
     @late_activities = Activity.debtors.select do |activity|
-      activity.impressionist_count(message: "mail",
-                                   start_date: activity.start) >= 4 && activity.is_payable
+      (Date.today.prev_occurring(:friday) - activity.is_payable_updated_at.to_date).to_i >= 21 && activity.is_payable
```
Riscky: maybe add a comment here explaining that this counts up to 4 weeks (last friday is more then 21 days ago)

Riscky: Can you also remove the superfluous comment on line 28?

app/models/activity.rb
```diff
@@ -49,6 +49,31 @@ def content_type
 
   attr_accessor :magic_enrolled_reservists
 
+  scope :late_unpayable, lambda {
```
Riscky, line 52: Is it possible to write this query as rails functions, or would that get too complex?

app/models/activity.rb
```diff
@@ -59,6 +84,11 @@ def name=(name)
     write_attribute(:name, name.strip)
   end
 
+  def is_payable=(is_payable)
```
Riscky, line 87: Can you add a comment that explains what this function does?

config/locales/admin.en.yml
```diff
@@ -98,6 +98,7 @@ en:
       date: Date
       day_total: Daily total
       days: Days
+      times: Sent mails
```
Riscky, line 101: the `times` key is not descriptive, can you change it to `sent_emails` or similar?

Riscky (changes requested):
Thanks for this PR, I'm really happy with the changes!
I have a few comments here and there, but generally this looks good :)

app/models/activity.rb
JustNireon: Get's a bit too complex for the time i have right now

app/models/activity.rb
```diff
@@ -59,6 +84,13 @@ def name=(name)
     write_attribute(:name, name.strip)
   end
 
+  # In case of the activity becoming payable we track the date so we can track how long people have been able to pay.
```
Riscky, line 87:
```suggestion
  # When the activity is set to payable we store the date so we can track for how long people have been able to pay.
```

Riscky (approved):
Looks LGTM to me

PR #1008, opened 2022-05-26
GPDR Mail

JustNireon:
Was checking if the system for GPDR was going to run correctly this summer. Everything seems functional except for the fact that mailchimp does net get updated when changing their status. To fix this we should just run a mailchimp job and they would receive their alumni tag. Additionally it would probably be preferable to also add an alumni list so even though alumnis keep their account they can easily toggle if they want to receive alumni mails.


Jelte-Akker (approved):
lgtm cutie

Jelte-Akker (approved):
lgtm

PR #1009, opened 2022-05-30
INSTALLING.md is now up-to-date

SilasPeters:
INSTALLING.md had some older information about how to set up Koala. For example, the command `bundle exec ./bin/webpack-dev-server` was never mentioned while it is necessary for Koala to run correctly.
In addition, some things were clarified a bit better and the overall flow to run Koala in the future is more explicitly explained.

```
  _____     ____
 /      \  |  o | 
|        |/ ___\| 
|_________/     
|_|_| |_|_|

My first PR! ^ This is Geoffry.
```

### DIFF
--- a/INSTALLING.md
+++ b/INSTALLING.md
@@ -70,18 +70,30 @@ docker-compose up -d
 
 PostgreSQL and Redis will now set itself up in the background, and will be available in
 a minute or so.
-You'll need to run the `docker-compose up` command again if you
-reboot your computer to start the database again.
-
+You'll need to run the `docker-compose up` command again (in a Nix-shell) if you
+reboot your computer to start the database again. If you like, you can run the
+container in the background by adding `-d` to the mentioned command.
 If you're already running a copy of PostgreSQL, you can use this too.
-You'll need to create a user with all privileges for the
-databases `koala-development` and `koala-test`, and set this in the `.env` file.
 
-There is an example file in the root of this repository called
+The previous command starts a process which should keep running while running Koala.
+In addition, the following dependency should also be kept running:
+
+```console
+bundle exec ./bin/webpack-dev-server
+```
+
+This command should also be executed from within a Nux shell.
+Thus, to run Koala correctly, these two processes should be running as well.
+
+
+Moving on, there is an example file in the root of this repository called
 `sample.env`. This file is a template for the actual configuration file
 `.env`, which sets some configuration values for Koala. Copy
 `sample.env` to `.env`, and edit it according to the
 instructions in the file.
+
+You'll need to create a user with all privileges for the
+databases `koala-development` and `koala-test`, and set this in the `.env` file.
 
 Once you're done, you can set up the database with this command (in the Nix shell):
 
--- a/INSTALLING.md
+++ b/INSTALLING.md
@@ -70,19 +70,19 @@ docker-compose up -d
 
 PostgreSQL and Redis will now set itself up in the background, and will be available in
 a minute or so.
-You'll need to run the `docker-compose up` command again (in a Nix-shell) if you
+You'll need to run the `docker-compose up` command again if you
 reboot your computer to start the database again. If you like, you can run the
 container in the background by adding `-d` to the mentioned command.
 If you're already running a copy of PostgreSQL, you can use this too.
 
-The previous command starts a process which should keep running while running Koala.
-In addition, the following dependency should also be kept running:
+The next command starts a process which enabled hot-reloading for the javascript and css.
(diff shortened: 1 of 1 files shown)


INSTALLING.md
```diff
@@ -70,19 +70,31 @@ docker-compose up -d
 
 PostgreSQL and Redis will now set itself up in the background, and will be available in
 a minute or so.
-You'll need to run the `docker-compose up` command again if you
-reboot your computer to start the database again.
-
+You'll need to run the `docker-compose up` command again (in a Nix-shell) if you
```
Riscky, line 73: `docker-compose` doesn't need to be run in a nix-shell

INSTALLING.md
```diff
@@ -70,19 +70,31 @@ docker-compose up -d
...
 
-There is an example file in the root of this repository called
+The previous command starts a process which should keep running while running Koala.
+In addition, the following dependency should also be kept running:
+
+```console
+bundle exec ./bin/webpack-dev-server
+```
+
+This command should also be executed from within a Nux shell.
```
Riscky, line 85:
```suggestion
This command should also be executed from within a Nix shell.
```

INSTALLING.md
```diff
@@ -70,19 +70,31 @@ docker-compose up -d
...
+You'll need to run the `docker-compose up` command again (in a Nix-shell) if you
+reboot your computer to start the database again. If you like, you can run the
+container in the background by adding `-d` to the mentioned command.
 If you're already running a copy of PostgreSQL, you can use this too.
-You'll need to create a user with all privileges for the
-databases `koala-development` and `koala-test`, and set this in the `.env` file.
 
-There is an example file in the root of this repository called
+The previous command starts a process which should keep running while running Koala.
+In addition, the following dependency should also be kept running:
```
Riscky, line 79: This should not strictly necessary when not working on the front-end.
I can imagine most people want hot reloading though, so maybe we should leave this as-is.

INSTALLING.md
```diff
@@ -70,19 +70,31 @@ docker-compose up -d
 
 PostgreSQL and Redis will now set itself up in the background, and will be available in
 a minute or so.
-You'll need to run the `docker-compose up` command again if you
-reboot your computer to start the database again.
-
+You'll need to run the `docker-compose up` command again (in a Nix-shell) if you
+reboot your computer to start the database again. If you like, you can run the
+container in the background by adding `-d` to the mentioned command.
 If you're already running a copy of PostgreSQL, you can use this too.
```
Riscky, line 76: This sentence belongs together with these lines:
> You'll need to create a user with all privileges for the
databases `koala-development` and `koala-test`, and set this in the `.env` file.

which you moved down a little bit.
I guess we can remove the lines, as most people that set up their own database will know how to set up things anyhow.

INSTALLING.md
Riscky: I also think we should try to set up a Nix environment with `foreman` or similar, which would give a oneline command to start the databases (Docker), the dev-server and Koala itself.
That would improve the developer first-time experience quite a bit, I'll look into this later.

Riscky (commented):
Thanks for your very first PR :)
I left a few comments, but when those are fixed this is good to go :+1:

INSTALLING.md
SilasPeters: I will remove them as I, someone fairly new to all this, didn't understand it, so taking that in account too I guess that most people don't have a use for it.

INSTALLING.md
SilasPeters: > This should not strictly necessary when not working on the front-end. I can imagine most people want hot reloading though, ...

I will mention what you said here, so that people can choose for themselves.

INSTALLING.md
```diff
@@ -71,13 +71,23 @@ docker-compose up -d
...
-You'll need to create a user with all privileges for the
-databases `koala-development` and `koala-test`, and set this in the `.env` file.
 
-There is an example file in the root of this repository called
+The previous command starts a process which should keep running while running Koala
+(unless you are part of the front-end, but it does enable hot-reloading. Choose for
+yourself). In addition, the following dependency should also be kept running:
```
Riscky, line 80:
```suggestion
The next command starts a process which enabled hot-reloading for the javascript and css.
If you're not working on the front-end, you can skip this step.
```

The text suggested that the previous command (docker-compose up) is optional, which is not the case.

Riscky (commented):
I have a single suggestion for the text to clarify the webpack command.
Before merging, can you also make squash your commits?

Jelte-Akker (approved):
lgtm!

PR #1010, opened 2022-05-30
Invoice fixes

JustNireon:
This PR fixes some current issues with the invoices created by exact online. I really dislike how this fixes some of the issues but it should suffice for now.


app/controllers/admin/payments_controller.rb
```diff
@@ -133,26 +133,37 @@ def create_invoice(csv, payments, payment_type, start_date, end_date = nil)
...
         # transaction description, VAT number, amount, cost_location ()
-        csv <<  if p.activity.group.nil? || Group.first.ledgernr.blank?
-                  ["", "1302", "#{ p.activity.name } - #{ p.member_id }", p.activity.VAT,
-                   p.currency, ""]
-                else
-                  ["", p.activity.group.ledgernr, "#{ p.activity.name } - #{ p.member_id }",
-                   p.activity.VAT, p.currency, p.activity.group.cost_location]
-                end
+        csv << if Activity.where(id: activity_id).description == "Lidmaatschap"
+                 # Dislike this way of doing it but need better ways to find membership activities
```
Riscky, line 137: inside this if-statement we know that it the `introduction` activity right?
We can set the ledger number to 8000 per default, and the vat to 0 then, removing the inner if-statement

app/controllers/admin/payments_controller.rb
```diff
@@ -133,26 +133,37 @@ def create_invoice(csv, payments, payment_type, start_date, end_date = nil)
...
+                 if p.activity.group.nil? || Group.first.ledgernr.blank?
+                   ["", "1302", "#{ p.activity.name } - #{ p.member_id }", p.activity.VAT,
+                    p.currency + payment.transaction_fee, ""]
+                 else
+                   ["", p.activity.group.ledgernr, "#{ p.activity.name } - #{ p.member_id }",
+                    p.activity.VAT, p.currency + payment.transaction_fee, p.activity.group.cost_location]
+                 end
+               elsif p.activity.group.nil? || Group.first.ledgernr.blank?
```
Riscky, line 145:
```suggestion
               elsif p.activity.group.nil? || p.activity.group.ledgernr.blank?
```

app/controllers/admin/payments_controller.rb
```diff
@@ -133,26 +133,37 @@ def create_invoice(csv, payments, payment_type, start_date, end_date = nil)
...
+                 ["", "1302", "#{ p.activity.name } - #{ p.member_id }", p.activity.VAT,
+                  p.currency, ""]
+               else
+                 ["", p.activity.group.ledgernr, "#{ p.activity.name } - #{ p.member_id }",
+                  p.activity.VAT, p.currency, p.activity.group.cost_location]
+               end
       end
     end
 
     payments.where(transaction_type: :checkout).group(:member_id).sum(:amount).each do |payment|
```
Riscky, line 155: Wouldn't it be better not to group by member_id here? if someone top ups multiple time per week, they will get multiple lines of csv, but that's fine.
It keeps this code a lot simpler in my opinion.

app/controllers/admin/payments_controller.rb
```diff
@@ -133,26 +133,37 @@ def create_invoice(csv, payments, payment_type, start_date, end_date = nil)
...
-      csv << ["", Settings.mongoose_ledger_number, "Mongoose - #{ payment[0] }", "9", payment[1],
-              ""]
+      transaction_costs = (payments.where(transaction_type: :checkout, member_id: payment[0]).count *
+      Settings.mongoose_ideal_costs)
+      csv << ["", Settings.mongoose_ledger_number, "Mongoose - #{ payment[0] }", "9",
+              payment[1] - transaction_costs, ""]
     end
 
-    activity_amount = payments.where(transaction_type: :activity).count
+    activity_amount = payments.count
```
Riscky, line 163: Don't we want to split out the transactions costs of activities and mongoose top-ups?
I can imagine we want to make Sticky pay the transactions costs for activities next year, but not for top-ups.

Riscky (changes requested):
Good to see work is done on this :)
I left a few comments and suggestions

app/controllers/admin/payments_controller.rb
JustNireon: Yeah i was thinking the same, but the possible issue could be if we change this for whatever reason we'd have to change the code but with this we could just change the group. I'm fine with either but your solution would be probably be fine aswell so i'll change it :)

app/controllers/admin/payments_controller.rb
JustNireon: Totally agree, was over complicating stuff here

app/controllers/admin/payments_controller.rb
JustNireon: Will split this up aswell!

app/controllers/admin/payments_controller.rb
Riscky: That is a fair argument.
The problem however is that the `Lidmaatschap` activity does not have group assigned when it's created:
https://github.com/svsticky/constipated-koala/blob/fffbcac81e339e2f47e3ea4bcae827b9a647ecaa/lib/tasks/admin.rake#L55-L62
so that should still be added then.

app/controllers/admin/payments_controller.rb
```diff
@@ -133,26 +133,37 @@ def create_invoice(csv, payments, payment_type, start_date, end_date = nil)
...
         # now create every transaction row with the following date: Blank, ledger account,
         # transaction description, VAT number, amount, cost_location ()
-        csv <<  if p.activity.group.nil? || Group.first.ledgernr.blank?
-                  ["", "1302", "#{ p.activity.name } - #{ p.member_id }", p.activity.VAT,
-                   p.currency, ""]
-                else
-                  ["", p.activity.group.ledgernr, "#{ p.activity.name } - #{ p.member_id }",
-                   p.activity.VAT, p.currency, p.activity.group.cost_location]
-                end
+        csv << if Activity.where(id: activity_id).description == "Lidmaatschap"
```
Riscky, line 136:
```suggestion
        csv << if activity_id == Settings['intro.membership']
```

Came across this in `lib/tasks/admin.rake`, seems like the way to go.

JustNireon: Thought about this aswell only issue being. In the hypothetical case you would create an invoice for introduction of previous years it would fail on this condition

Riscky: I don't know if that is a big issue;
If the old membership activity does not have a group, it will become a 1302, which is not wrong, just less informative.
If the old membership activity does have a group, it should be set correctly anyhow.

Riscky (approved):
With the changes this now LGTM :)
If you have tested this, this is ready to merge :+1:

Riscky (approved):
LGTM!

PR #1011, opened 2022-05-30
remove old, broken status mail code

Riscky:
Title says all.
The mailing code is called nowhere, doesn't work, and has been replaced by the `status.rb` mailing.
The mailing did not have an accompanying html view.


Raymonzut (approved):
LGTM, easy deletion of unused code

PR #1014, opened 2022-06-07
Prepare oauth for undead mongoose

Phavion:
Adds an is_admin claim that lets undead mongoose know whether it should create a superuser or not.


Riscky (approved):
LGTM

PR #1015, opened 2022-06-08
Sac points tracking system:tm: 

### LINKED ISSUE #918: Add SAC Feature to Koala
We want to be able to keep record of the amount and what commissions a person has done AND certain activities.

Each commission and some type of activities will award a certain amount of points, that overview will be linked here (Keep in mind that the points are Subjec to change, this change will happen once though.. so hardcoding them isnt a problem):
[soon incoming]

Per user we want to keep track of the point categories, below a short task list of the key points we want implemented:

- [ ] Add feature to add SAC activity category to activities.

- [ ] Keep track of the SAC activities, and the points, the user has.

- [ ] Be able to calculate how many SAC commissions, and the corresponding amount of points, the user has

- [ ] Give a SAC overview in the profile of the user

- [ ] (last task) Make a download button which gives a pdf with the overview

stickyPiston:
finally (will fix #918)


JustNireon:
Commissions are missing entirely from this PR right now. Each commission should have their own SAC points for being a member of the commission. This was a quick review will do a bit more indepth later on

app/views/admin/members/show.html.haml
```diff
@@ -253,6 +253,31 @@
...
+            = link_to I18n.t("admin.sac.download"), member_sac_path(@member.id)
+          %table.table.table-striped#sac
+            %thead
+              %tr
+                %th= I18n.t('admin.activities.single')
+                %th= I18n.t('admin.sac.category')
+                %th= I18n.t('admin.sac.title')
+            %tbody
+              - total = 0
+              - @activities.select{ |ac| ac.sac_category? or ac.participants.where(:member => @member).first.sac_points?}.each do |activity|
```
Riscky, line 268: Can you move this logic to the controller?
It is unavoidable to have a loop here, but a query doesn't really mix with well with a view in my opinion.

yarn.lock
```diff
@@ -0,0 +1,4854 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
Riscky, line 1: Please do not check in lockfiles.
All dependency versions are locked by Nix, our JavaScript dependencies in `nix/node_modules.nix` specifically.
It is fine to use yarn locally, although this should not be necessary.
If you're using yarn because something in the Nix setup is broken, let me know.

config/initializers/sac.rb
```diff
@@ -0,0 +1,6 @@
+SAC_CATEGORIES = [
```
Riscky, line 1: Configuration like this should live in the `config` directory.
I think `config/application.rb` is the best location, as comparable configuration also lives there.
There currently is no way to edit the categories as an admin, but since this won't change too often, it is fine.
Make sure the categories are filled with something sensible before merging though, as this configuration will be used in production.
@Jelte-Akker should be able to help you out with the relevant categories and points.

Riscky (changes requested):
Great to see this PR!
In general it looks really good, well done :)
I left a few comments I'd like to see you address, but apart from that this looks quite ready to merge.
Let me know if you have questions about anything I wrote.

Riscky:
> Commissions are missing entirely from this PR right now. Each commission should have their own SAC points for being a member of the commission. This was a quick review will do a bit more indepth later on

Depending on the exact implementation details of the committee points, this might be something for a later PR.
If the amount of committee points is equal to the amount of committees, it should be as simple as adding
`member.groups.where(category: :committee).count`, and that could be added in this PR.
Something more involved should be moved to a later PR in my opinion.
The reasoning being that it would be fairly easy for the board to calculate the committee points by hand, while waiting to merge this PR means that way more work has to be done by the board.
@Jelte-Akker could you give more details here on how this should be calculated?

config/initializers/sac.rb
Riscky: Apparently the categories and points are linked in the issue, so you can use that.

yarn.lock
stickyPiston: Ah I hadn't actually noticed this, you got a good point. I'll remove it from the PR

stickyPiston:
All requested changes are implemented and this PR only needs the sac categories to be filled out. Waiting for further actions now

PR #1018, opened 2022-06-20
read OIDC signing key from env var

Riscky:


Riscky:
@Mstiekema can you still approve this?

Mstiekema (approved):

PR #1019, opened 2022-06-22
Reservist sign-out button color change

SilasPeters:
Now, the sign-out color is a more darker (scarier) orange so that the user can easily distinguish signed-in activities from others.
Previously, it was the same default orange used as a warning color.

This idea is brought to you by Jack.


Riscky (approved):
LGTM

PR #1033, opened 2022-10-26
Fix enroll reservists

JustNireon:


SanderHageman:
The issue is a bit 'vague' to me, and I haven't noticed it working incorrectly. Do the changes resolve the issue now?

JustNireon:
> The issue is a bit 'vague' to me, and I haven't noticed it working incorrectly. Do the changes resolve the issue now?

Not sure, we have had issues and this was a hopeful fix, but the issue overall seems to occur very rarely

SilasPeters (approved):
I see no harm in this change, so I suggest we just merge this.

@Siem2l would you like to do the honor, or shall I do it?

PR #1035, opened 2022-10-26
fix csv export for treasurer

Siem2l:
Don't kjnow if this fixes it should check this later


app/controllers/public/home_controller.rb
```diff
@@ -52,14 +52,13 @@ def create
 
       transaction = Payment.new(
         description: I18n.t("form.introduction", user: @member.name),
-        amount: total,
+        amount: total - Settings.mongoose_ideal_costs, # Transaction costs are paid by the association
```
SanderHageman, line 55: This would have been nice to have merged before the start of the summer 😓

app/controllers/admin/payments_controller.rb
```diff
@@ -151,8 +151,13 @@ def create_invoice(csv, payments, payment_type, start_date, end_date = nil)
       csv << ["", Settings.mongoose_ledger_number, "Mongoose - #{ payment.member_id }", "9",
               payment.amount - Settings.mongoose_ideal_costs, ""]
     end
+    # minus lidmaatschap payments
+    activity_amount = payments.where(transaction_type: :activity).count - payments.where(
+      transaction_type: :activity,
+      transaction_id: Settings['intro.membership']
```
SanderHageman, line 157: Since the name of the activity is forced to "Lidmaatschap" anyway, consider filtering on that to somewhat resolve your todo

Maybe also put the "Lidmaatschap" in a variable somewhere instead of hardcoding it

SanderHageman (changes requested):
Super useful stuff <3
As you noted yourself in the TODO, maybe look at my suggestion? Cheers,

app/controllers/public/home_controller.rb
JustNireon: Uhu

app/controllers/admin/payments_controller.rb
JustNireon: Don't know if i like the practice of filtering on a string that's so easiliy configured by an admin, the other consideration is you could also just change the activity id in settings

SanderHageman (approved):
lgtm 🚀

PR #1051, opened 2023-10-06
Activy URL is now displayed in the appointment description

SilasPeters:
When adding an activity to your calendar through the new link (as mentioned in #1045), the link to the activity is not included in the description of the activity.

I can't run koala locally as of now, but I believe this syntax might be correct. If not, please correct it :)

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -365,7 +365,7 @@ class Activity < ApplicationRecord
     return nil if start.nil? || self.end.nil?
 
     loc = I18n.locale if loc.nil?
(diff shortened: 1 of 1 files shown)


SilasPeters:
@stickyPiston why does the linter complain? I intentionally don't want spaces inside my string interpolation. I use interpolation because the linter forced me to do so earlier :(

P.S. Could you run this locally to see if these changes do what I expect them to?

stickyPiston:
Everything looks good and works like it should, time to merge it!

stickyPiston (approved):

PR #1070, opened 2023-10-29
fix treasurer whatsapp payment reminder

SanderHageman:
- now uses newer wa.me api
- resolve newline before colon
- display message in member-language instead of admin-language


HugoPeters1024 (approved):
LGTM

PR #1078, opened 2023-11-12
Automatically add opened PRs to project

SilasPeters:
This already happens to issues. Now PR's should be included as well. Not tested, but should work.


SilasPeters:
Well, it has already proven itself!

stickyPiston (approved):
よくやった

PR #1080, opened 2023-12-12
Button to add committe to activity :)

PrinceMordred:
Uses the participant.enroll method to enroll each committee member (in activity.js)


Mstiekema (approved):
LGTM :shipit:

PR #1082, opened 2023-12-12
Removed the html required to interact with the controller

PrinceMordred:
Remove the ability to add funds or cards from koala admin portal


SilasPeters:
It was quite ready to merge lol, but it has already been done in other PRs meanwhile

PR #1084, opened 2024-01-18
Changed calendar end time to 11:59pm when one isn't set

### LINKED ISSUE #1083: Add to calendar: Incorrect end date for activities spanning whole day(s)
## Current Behaviour

When clicking on the link to add an activity to your calendar, (and when the activity spans over multiple days,) and when no specific end time is specified for the final day (assuming koala stored the activity as lasting the whole day), the timestamp filled in your agenda is 00:00 on the start of the final day.

The consequence is, that when an activity is from 1-3 Jan, but no end time is specified (just 3 Jan), then the agenda displays the activity only on 1-2 Jan, as it ends at 3 Jan 0:00.

## Expected Behaviour

The timestamp of 3 Jan is 23:59, or even better, the calendar knows 'whole days' are specified and no timestamps have to be supplied.

## Notes

I believe the bug is that we specify no timestamp, but also not that the activity spans over whole days - without specific timestamps.

This bug might also influence activities which do not span over multiple days, but just do not have a specific end time

stickyPiston:
fixes #1083

### DIFF
--- a/app/models/activity.rb
+++ b/app/models/activity.rb
@@ -371,7 +371,8 @@ class Activity < ApplicationRecord
     uri_name = URI.encode_www_form_component(name)
     uri_description = URI.encode_www_form_component(description)
     uri_location = URI.encode_www_form_component(location)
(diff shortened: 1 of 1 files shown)


SilasPeters (approved):
LGTM

PR #1086, opened 2024-01-24
Increased char limit to member notes

SilasPeters:
This is not tested right now. Its late, and I have connection errors...


stickyPiston (approved):
頑張った！

Issue #1088, opened 2024-02-08
Hook up mongoose credit view to new mongoose backend


stickyPiston:
The credit view on the homepage stopped working for quite a while now, so we can
- fix it, if there is an api endpoint in mongoose we can use;
- temporarily remove it so it avoids confusion;
- or something else if someone has a suggestion.

PR #1091, opened 2024-02-19
Remove mongoose balance from the member home page

TobiasDeBruijn:
The balance is not correct, and serves no purpose since the decoupling of Mongoose from Koala


stickyPiston:
fixes #1088

stickyPiston (approved):
頑張った！

PR #1098, opened 2024-03-13
Update link to 'Photos' site to Chroma instead of Pxl

TobiasDeBruijn:


SilasPeters:
While this is a personal preference, I suggest to keep `fotos.svsticky.nl` as this is way more recognizable and rememberable than `chroma.svsticky.nl`. I would keep both pxl and chroma for the time being, and let fotos.svsticky.nl redirect to chroma. Or even better, if we do just change the OAuth urls, let chroma ba accessible by `fotos.` instead.

TobiasDeBruijn:
While this PR doesn't address that, rather it just changes the URL in the link in Koala, it is something that needs to be throught about. Jack or Sander, opinions?

SilasPeters:
Well, if we do choose to keep photos.svsticky.nl pointing to chroma, then koala should just link to photos and not chroma. I believe this PR changes koala to link to chroma directly

PR #1101, opened 2024-05-21
feat: personalised ical feed

### LINKED ISSUE #1037: Generate and maintain user-specific calendars containing the corresponding users' activities
Currently, constipated-koala allows you to view your activities in two ways: by subscribing to a calendar containing all activities of Sticky, or by using the soon-to-be button which generates an .ics file for a specific activity. The latter functions as an alternative to the global calendar, because it clutters the users calendar. However, to use this alternative you have to manually download every activity you sign up for.
Combining the best of both worlds, this issue suggests that - using the google API - we create and manage personal calendars containing all activities for which a user signed up for. This way, you have the automatisation of the original global calendar, but the personal experience of the .ics files. The user does not have to interact with his calendar, yet all his personal activities show up in his calendar (given that he subscribed to his personal sticky calendar).

My suggestion is that, next to the button allowing you to subscribe to the global calendar, a new button is added which allows you to subscribe to your personal activities calendar. Potentially, the calendar isn't generated (and maintained) until you first click that button to reduce overhead.

SilasPeters:
This PR introduced a new button, 'Copy personalised Webcal link`, which copies the link to your ical feed which any calendar client can use to visualise the activities where you are enrolled.

It also adds clarification on the difference between the ical feed and the ical templates

This will close #1037


lib/icalendar_helper.rb
```diff
@@ -0,0 +1,40 @@
...
+  def self.activity_to_event(activity, locale)
+    event = Icalendar::Event.new
+    event.uid = activity.id.to_s
+    event.dtstart = activity.start_date
+    event.dtend = activity.end_date
+    event.summary = activity.name
+    event.description = activity.description_localised(locale)
+    event.location = activity.location
+    return event
+  end
```
stickyPiston, line 15: I reckon the activity model class might be a better place for this method

lib/icalendar_helper.rb
```diff
@@ -0,0 +1,40 @@
...
+  # Combines zero or more Icalendar events into an iCalendar abstract object
+  def self.create_calendar(events, locale)
+    calendar = Icalendar::Calendar.new
+    calendar.x_wr_calname =
+      case locale
+      when :nl
+        "Sticky Activiteiten"
+      else
+        "Sticky Activities"
+      end
```
stickyPiston, line 26: This seems overly verbose, i reckon it would be better to put these strings in the i18n file and reference that

app/controllers/api/calendars_controller.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  # Not exposed to API directly, but through #show
+  def create_personal_calendar
+    @locale = I18n.locale
+
+    # Convert activities to events, and mark activities where the member is
+    # is enrolled as reservist
+    @reservist_activity_ids = @member.reservist_activities.ids
+    events = @member.activities.map do |a|
+      a.name = "[RESERVIST] #{ a.name }" if @reservist_activity_ids.include?(a.id)
```
Siem2l, line 43: Reservist is niet localized :(

app/models/member.rb
```diff
@@ -203,6 +204,12 @@ def groups
     return groups.values
   end
 
+  # We cannot enforce that a member is created with a calendar_id, but we can enforce that
```
Siem2l, line 207: This sentence is a contradiction, might want to rewrite it because now I'm not sure if this sentence makes sense.

config/locales/en.yml
```diff
@@ -35,7 +35,7 @@ en:
         description_nl: Dutch description
         end_date: Enddate
         end_time: Endtime
-        google_event: Add to calendar
+        google_event: Copy once to calendar
```
Siem2l, line 38: Not entirely sure the context, but don't think sentence makes sense either.

config/locales/nl.yml
```diff
@@ -35,7 +35,7 @@ nl:
         description_nl: Nederlandse beschrijving
         end_date: Einddatum
         end_time: Eindtijd
-        google_event: Voeg toe aan je agenda
+        google_event: Kopiëer eenmalig naar kalender
```
Siem2l, line 38: Feel like the instruction of only copying once should be more elaborate and not incorporated into a title. Pretty sure people will not understand this.

lib/icalendar_helper.rb
```diff
@@ -0,0 +1,40 @@
...
+    return event
+  end
+
+  # Combines zero or more Icalendar events into an iCalendar abstract object
+  def self.create_calendar(events, locale)
+    calendar = Icalendar::Calendar.new
+    calendar.x_wr_calname =
+      case locale
+      when :nl
+        "Sticky Activiteiten"
```
Siem2l, line 23: Localize :(

Siem2l (changes requested):
Few comments

lib/icalendar_helper.rb
SilasPeters: Agreed

config/locales/nl.yml
SilasPeters: Not sure how to fix this. There's only limited space, and as it's a hyperlink it would be weird to just add a button next to it for more info. I guess I will add one of those little question marks in a circle

SilasPeters: Now that I think of it, I don't think that adding more information might be necessary.

The whole thing I am trying to convey, is that when you add the activity to your calendar through that hyperlink, that it does not 'subscribe' your calendar to the activity, but simply copies it over as-is.

When the text says 'copy once to calendar, and the opened webpage displays a filled in template for that single activity, my hopes are that the user will understand that he should not rely on that activity in his calendar to be updated when koala updates.

What do you think?

SilasPeters:
I now realise that there are also activities where you cannot sign up for. This might raise confusion, for you bought a pretix ticket but the activity does not appear in your calendar.

I should add some clarification on what exactly the personalized calendar feed is, or add a checkbox next to every activity to include the activity in your calendar, where it is automatically checked when you sign up.

config/locales/nl.yml
SilasPeters: I believe things are clear, now that I give an explanation on the fact that koala is only aware of the activities in koala

db/migrate/20240521180025_add_calendar_id_to_member.rb
```diff
@@ -0,0 +1,18 @@
...
+    add_column :members, :calendar_id, :uuid
+
+    # Populate existing records with a random UUID
+    Member.reset_column_information
+    Member.find_each do |record|
+      record.update_columns(calendar_id: SecureRandom.uuid)
+    end
+
+    # Add an index to the UUID column, because why not
+    add_index :members, :calendar_id, unique: true
```
stickyPiston, line 12: Nice touch, good to enforce uniqueness of the calendar id

config/locales/en.yml
```diff
@@ -33,9 +33,10 @@ en:
         description: Description
         description_en: English description
         description_nl: Dutch description
+        disclaimer: Data on this activity may be outdated, as it was addes as a one-time copy of the information given at that time. Up-to-date info can be found on Koala.
```
stickyPiston, line 36: Is this translation used anywhere still?

config/locales/members.en.yml
```diff
@@ -16,6 +16,11 @@ en:
         unenroll: Unenroll
         update_info: Update info
       back_to_overview: Back to activity overview
+      calendar:
+        confirm_understand_icalendar:
+          confirm: I understand
+          text: Koala will maintain the icalendar feed you just copied, at all times. However, koala is only aware of activities for which you are enrolled through koala, and not trough some other platform (like Pretix or a Google form). Remember to add those activities to your calendar manually.
+          title: Warning
```
stickyPiston, line 23: Clear explanation, good that you've added a warning on clicking the button

config/locales/members.nl.yml
```diff
@@ -26,6 +31,7 @@ nl:
       index:
         activities_calendar: Activiteitenkalender
         copy_ICS: Kopieer Webcal link
```
stickyPiston, line 33: The dutch translation doesn't include the "for all activities part".

stickyPiston (dismissed):
頑張ったね！

stickyPiston:
Also on a separate note: it might be nice to mention how to import the personalised calendar into one's own calendar app somewhere. But i'd love to see this feature get merged as soon as possible first.

config/locales/en.yml
SilasPeters: It is! In the description of an activity

config/locales/members.nl.yml
SilasPeters: Fuck

stickyPiston (approved):
また頑張ったね！

PR #1103, opened 2024-05-31
Add note type for activities

Siem2l:
I wondered how long it would take to implement this without altering too much in the database. I don’t think this is necessarily the best way to implement it, but it seems fully functional at least.

Only the translations are missing, I think.
image

image

### DIFF
--- a/app/controllers/members/participants_controller.rb
+++ b/app/controllers/members/participants_controller.rb
@@ -28,7 +28,8 @@ class Members::ParticipantsController < ApplicationController
     end
 
     @member = Member.find(current_user.credentials_id)
-    @notes = params[:par_notes]
+    @notes =  params[:par_notes]
+      
     reservist = false
 
     # Deny if already enrolled
@@ -190,6 +191,7 @@ class Members::ParticipantsController < ApplicationController
     )
 
     if @activity.notes.blank? || params[:par_notes].present?
+      updated_notes = params[:par_notes]
       @enrollment.update(notes: params[:par_notes])
       @enrollment.save
       render(status: :accepted, json: {
--- a/app/javascript/src/admin/activities.js
+++ b/app/javascript/src/admin/activities.js
@@ -540,3 +540,91 @@ function posterHandlers() {
     $(this).find('button[type="submit"].wait').addClass("disabled");
   });
 }
+
+$(document).on('click', '#add-notes-option', function() {
+  var optionIndex = $('#notes-options .col-md-6').length;
+  var optionHtml = '<div class="col-md-6">' +
+    '<div class="input-group mb-3">' +
+    '<input class="form-control" type="text" name="activity[notes_options][]">' +
+    '<div class="input-group-append">' +
+    '<button class="btn btn-sm btn-danger delete-option" type="button">' +
+    '<i class="fa fa-fw fa-minus"></i>' +
+    '</button>' +
+    '</div>' +
+    '</div>' +
+    '</div>';
+  $('#notes-options').append(optionHtml);
+});
+
+$(document).on('click', '.delete-option', function() {
+  $(this).closest('.col-md-6').remove();
+});
+
+$(document).on('change', '#activity_notes_input_type', function() {
+  var notesInputType = $(this).val();
+  var notesContainer = $('#notes-container');
+
+  if (notesInputType === 'text') {
+    var notesValue = $('#notes-options input').map(function() {
+      return $(this).val();
+    }).get().join('\n');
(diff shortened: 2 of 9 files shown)


app/javascript/src/members/activities/activity.js
```diff
@@ -184,12 +197,29 @@ Object.defineProperties(Activity.prototype, {
      */
     value: function (method) {
       var activity = this;
+      var par_notes;
+      var notes = [];
+      console.log(activity.has_notes_checkboxes);
+      if (activity.has_notes_checkboxes()) {
```
SanderHageman, line 203: I also know little JS, however, does it not have switch cases?

SanderHageman: Is there any way to define the enum in 1 place like in `activity.rb` and reference it from the jabbascript?

app/controllers/members/participants_controller.rb
```diff
@@ -28,7 +28,8 @@ def create
     end
 
     @member = Member.find(current_user.credentials_id)
-    @notes = params[:par_notes]
+    @notes =  params[:par_notes]
```
SanderHageman, line 31: Why did the linter add a double space? Linters 🤷

app/javascript/src/members/activities/activity.js
```diff
@@ -184,12 +197,29 @@ Object.defineProperties(Activity.prototype, {
...
+        console.log(this.notes.find(":checked"));
```
SanderHageman, line 204: remove log or cache `this.notes` call

SanderHageman (changes requested):
Minor changes

Could only partially check because I can't read html/css and barely understand js and haml.

db/migrate/20240530151252_add_notes_input_type_to_activities.rb
```diff
@@ -0,0 +1,5 @@
+class AddNotesInputTypeToActivities < ActiveRecord::Migration[6.1]
+  def change
```
SilasPeters, line 2: If rolled back, comments will still be filled with string containing special syntaxes, I suppose? Perhaps these should be processed or replaced. I don't know what the norm is.

app/javascript/src/admin/activities.js
```diff
@@ -540,3 +540,110 @@ function posterHandlers() {
...
+        '<div class="col-md-6">' +
+        '<div class="input-group mb-3">' +
+        '<input class="form-control" type="text" name="activity[notes_options][]">' +
+        '<div class="input-group-append">' +
+        '<button class="btn btn-sm btn-danger delete-option" type="button">' +
+        '<i class="fa fa-fw fa-minus"></i>' +
+        "</button>" +
+        "</div>" +
+        "</div>" +
+        "</div>";
```
SilasPeters, line 633: Perhaps add some indentation here?

app/javascript/src/members/activities/activity.js
```diff
@@ -184,12 +197,29 @@ Object.defineProperties(Activity.prototype, {
      */
     value: function (method) {
       var activity = this;
+      var par_notes;
+      var notes = [];
+      console.log(activity.has_notes_checkboxes);
```
SilasPeters, line 202: Shouldn't this be removed?

SilasPeters (changes requested):

app/javascript/src/members/activities/activity.js
Siem2l: Don't think so, atleast not trivally

db/migrate/20240530151252_add_notes_input_type_to_activities.rb
Siem2l: The special syntax is , or \n seperated. Don't think we should do any special processing to ensure backwards compatibility

app/javascript/src/members/activities/activity.js
Siem2l: Yes

app/javascript/src/admin/activities.js
Siem2l: Might be preferable, however, an even better approach would be to eliminate the need for generating HTML within the JavaScript code altogether.  

Make a HAML partial file as a template for the option menu.
In the _edit haml file render the template but keep it hidden with css attributes
Update the code that appends the options HTML to use the cloned template.
Modify the code that sets the input values to target the cloned elements and make it visible
Add an event listener for the "Add Option" button to clone and append the template.
Add an event listener for the "Delete Option" button to remove the corresponding option.

PR #1106, opened 2024-07-10
Remove checkout transactions

TobiasDeBruijn:
This PR removes the remaining UI elements regarding checkout transactions.
Affects only the admin page:
- Home
- Payments page

Checkout is no longer part of Koala, and now lives as a separate application.


SanderHageman (approved):
Would be nice to add an issue (or fix with this PR) to remove the dead code in the ruby code; the "apps" for checkout and transactions can also be removed from koala

TobiasDeBruijn:
Fair yeah. I'll delegate that to another PR to at least have the UI cleaned up for now, I'll go with a shovel for the rest of the backend code at a later date then.

PR #1107, opened 2024-07-22
Add the time an activity sign-up opens to the activities page

### LINKED ISSUE #1100: Feat: Activity sign up time
# Idea

There is currently no indication for members when the sign up opens for an activity.
When creating the activity this field is present, we should probably show this on the activity for members.

TobiasDeBruijn:
Fixes #1100

### DIFF
--- a/app/views/members/activities/partials/_view.html.haml
+++ b/app/views/members/activities/partials/_view.html.haml
@@ -46,7 +46,7 @@
           = activity.end_time.strftime("%H:%M")
       <br/>
       - if activity.open_date != nil && !activity.open?
-        %strong= I18n.t("activerecord.labels.activities.opens_at")
+        %strong= I18n.t("activerecord.labels.activities.opens_from")
         = l(activity.open_date, format: "%A %d/%m/%Y")
         - if !activity.open_time.nil?
           = l(activity.open_time, format: "%H:%M")
--- a/app/views/members/activities/partials/_view.html.haml
+++ b/app/views/members/activities/partials/_view.html.haml
@@ -45,6 +45,12 @@
           \-
           = activity.end_time.strftime("%H:%M")
       <br/>
+      - if activity.open_date != nil && activity.open_date > Date.today
+        %strong= I18n.t("activerecord.labels.activities.opens_at")
+        = l(activity.open_date, format: "%A %d/%m/%Y")
+        - if !activity.open_time.nil?
+          = l(activity.open_time, format: "%H:%M")
+        <br/>
       %strong= I18n.t("activerecord.attributes.activity.price")
       - if !activity.price.nil? and activity.price != 0
         &euro;
--- a/app/views/members/activities/partials/_view.html.haml
+++ b/app/views/members/activities/partials/_view.html.haml
@@ -45,7 +45,7 @@
           \-
           = activity.end_time.strftime("%H:%M")
       <br/>
-      - if activity.open_date != nil && activity.open_date > Date.today
+      - if activity.open_date != nil && !activity.open?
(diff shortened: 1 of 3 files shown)


app/views/members/activities/partials/_view.html.haml
```diff
@@ -45,6 +45,12 @@
           \-
           = activity.end_time.strftime("%H:%M")
       <br/>
+      - if activity.open_date != nil && activity.open_date > Date.today
```
SanderHageman, line 48: A quick google suggests to me that `Date.today` in rails shouldn't really be used, [please validate](http://winstonyw.com/2014/03/03/time_now_vs_time_zone_now/). I haven't checked which time/function is used for the check to see if someone can register for an activity but these need to be the same and I would suggest creating a helper function in ruby

app/views/members/activities/partials/_view.html.haml
```diff
@@ -45,6 +45,12 @@
...
+        %strong= I18n.t("activerecord.labels.activities.opens_at")
+        = l(activity.open_date, format: "%A %d/%m/%Y")
+        - if !activity.open_time.nil?
```
SanderHageman, line 51: I'm unsure about the interactivity between ruby and haml but the activity model also contains a `when_open` method which includes both time and date. Current logic suggests to me that the open date will be visible for the whole day the activity opens; is that intended?

SanderHageman (changes requested):
Some questions

app/views/members/activities/partials/_view.html.haml
TobiasDeBruijn: We use `Date.today` everywhere for activities, so I think it'd be best to stick with that?

app/views/members/activities/partials/_view.html.haml
TobiasDeBruijn: Good catch, it is not.

Looking into it further, there is actually a function on Activity: `open` (`:: bool`), so I'll just use the inverse of that.

config/locales/en.yml
```diff
@@ -382,6 +382,7 @@ en:
     labels:
       activities:
         date: Date
+        opens_at: Signup from
```
SilasPeters, line 385: Although 'Signup from' is fine, I would like to propose two alternatives to see what you prefer.

"Enrollable from", the confirmation message when you signed up for an activity says "Succesfully _enrolled_ for this activity" and this is more consistent with that. I am not certain whether this is better.

"(Admissions) Open(s) at", not for a particular reason, just reads better imo.

SilasPeters (changes requested):
I like these changes, and they are all idiomatic. However I suggest that when an activity will be enrollable at some point in the future, that the 'enroll' button is not hidden but grayed out. This better conveys the message of "you can/must enroll to participate!"

This can be another PR, but I suggest we do it now before we forget. Hence the "requested changes."

config/locales/en.yml
SilasPeters: Also, shouldn't "open_at" be "open_from"? I don't know if that is more idiomatic

app/views/members/activities/partials/_view.html.haml
SanderHageman: consistency 🫠

SanderHageman (dismissed):

config/locales/en.yml
TobiasDeBruijn: Good point. I'll change the key to `open_from` and change the text to `Enrollable from`.

The button will be another PR, it isn't relevant here :)

SilasPeters (approved):
Ready to roll

PR #1112, opened 2024-07-29
Update background for intro.svsticky.nl

TobiasDeBruijn:
The old one was due to be replaced.

PR #1114, opened 2024-07-30
Update intro landing page style

SpookyBoy99:
The form elements on the registration looked horribly broken and the background color of the 'What does Sticky do?' section had to little contrast to read comfortably. The background also requires an update, but that is being worked on in #1112.

**What is missing:**
- [x] Translations for client side error messages
- [x] A fix that will automatically hide the payment section on page load when a masters study is chosen


TobiasDeBruijn (dismissed):

TobiasDeBruijn (dismissed):

app/views/public/home/index.html.haml
SilasPeters, line 1: The text in the header bar moves during scrolling: boldness of the active item is taken into account when translating the items (if I am using the correct jargon).

app/javascript/src/intro.js
```diff
@@ -4,66 +4,89 @@ import I18n from "./translations.js";
...
 
-      // field is valid if member is 18 years or if it's not empty
-      return birthDate < maxDate || value;
+function setup_form_validation() {
+  jQuery.validator.addMethod(
+    "phone",
+    function (value) {
+      return (
+        value.trim().length === 0 ||
+        /^\+?(?:[0-9] ?){6,14}[0-9]$/.test(value.trim())
```
SilasPeters, line 86: This regex matches a + sign. Is `value` here the full +31 6 69420 phone number, or just the text that was given by the user?

app/javascript/src/intro.js
```diff
@@ -4,66 +4,89 @@ import I18n from "./translations.js";
...
+      );
     },
-    I18n.t("form.phone_required_underage"),
+    "Invalid phone number",
```
SilasPeters, line 89: Did you forget localisation here?

config/locales/public.nl.yml
```diff
@@ -15,14 +15,18 @@ nl:
     privacy: Privacy
   form:
     bank: Bank
+    digits_field: Dit veld accepteert alleen cijfers
     finish_signup: Inschrijving afronden
     general: Algemene gegevens
     introduction: Introductie %{user}
+    invalid_email: Emailadres is niet geldig
```
SilasPeters, line 22: E-mail adres toch?

app/javascript/src/intro.js
SilasPeters, line 1: De form validation lijkt de volgende fouten te hebben, gebaseerd op de errors die gegeven worden:
- Alle velden die text toelaten kunnen volledig whitespace zijn
- Mijn telefoonnummer mag +31 hebben vooraan
- Mag mijn hoofdtelefoonnummer (niet voor calamiteiten) express leeg zijn?
- Ik lijk niet een studie te hoeven kiezen

Ik heb nooit kunnen voldoen aan alle huidige form validations, omdat de 'Bank' field geen opties bevat, welke niet leeg mag zijn. Ik hoop dat deze test alsnog relevant is.

SilasPeters, line 1: Kan de date field naar `dd/mm/yyyy`? Puur mijn voorkeur

SilasPeters (changes requested):
I like the yellow, although orange is Sticky's official color right? (Heavily biased, orange is my favorite color.)

Aside from this, I have some questions and a nitpick or two

app/javascript/src/intro.js
SilasPeters: During testing I could enter `+31 6 123456`. Will this be send to the server as `+31 +31 6 123456`? I could not find this that quickly (not a front-ender)

SpookyBoy99: The + sign is optional due to bootstrap; it will automatically add +31 (or whatever is selected) if it is not in the entered by the user. The original regex made it mandatory which messed with the system of the form. In my testing it works as expected. That being said, the server also checks the phone number and that is leading, so it is better if the server is more strict that the front-end.

app/views/public/home/index.html.haml
SpookyBoy99: Your question being?

app/javascript/src/intro.js
SpookyBoy99: I have changed the date field, depending on your localization this should be default.

app/views/public/home/index.html.haml
SilasPeters: Oh I assume that's unintentional. The old version does not have that the text moves around depending on which one is bold. My opinion is that it looks better if it doesn't, but if it's my opinion against yours just leave it

config/locales/public.nl.yml
SilasPeters: Oof daar gaat mijn middelbare kennis

app/views/public/home/index.html.haml
SpookyBoy99: Oh I now get what you mean; the old version did indeed not make the current section bold in the navbar, but I thought the contrast and clarity was a bit lacking with bootstraps implementation so I decided to make te active item bold. This does indeed introduce the side effect of making the items a bit 'jumpy', but I don't think it is much of a problem. It is fixable, but only by giving the menu items a fixed width which gives it a weird appearance and will give problems in the future if more items are added. I could remove the boldness again, but is ultimately a trade-off.

app/javascript/src/intro.js
SpookyBoy99: Good points, I'll have to check this tomorrow. The 'Bank' field will always be empty on local deployments, however, masters don't have to pay so that way you can bypass this problem.

app/views/public/home/index.html.haml
SilasPeters: Keep the boldness! If it allows for better code, just leave the jumpiness

config/locales/public.nl.yml
SpookyBoy99: Dit is exact de link die ik zelf heb gebruikt om op te zoeken hoe het alweer moest, je bent niet de enige ;p

SpookyBoy99:
I have replaced the jquery-validation plugin scripts in `vendor/` with an updated npm package to fix the issue of just white space satisfying the constraints in required fields. I do think this requires an update to cachix and I actually expected the build to fail, but that runs fine. It might cause issues when deploying though.

app/javascript/src/intro.js
SpookyBoy99: Fixing the last item proved to be more of a challenge then I expected. It should work fine now and also be reasonably future proof, unless the study fields are removed altogether without updating `intro.js`, but would not be wise anyhow.

app/javascript/src/intro.js
```diff
@@ -78,96 +79,142 @@ function setup_background_parallax() {
...
+      required: {
+        depends: function () {
+          // get max birth date to be 18 years old, ignoring time
+          const adultBirthDate = new Date();
+          adultBirthDate.setUTCHours(0, 0, 0, 0);
+          adultBirthDate.setFullYear(adultBirthDate.getFullYear() - 18);
+
+          const birthDate = new Date($("#member_birth_date").val());
+
+          return birthDate > adultBirthDate;
```
SilasPeters, line 123: Nice change

SilasPeters (approved):
> I do think this requires an update to cachix and I actually expected the build to fail

I manually did every step from sadserver I could do locally, and everything worked. The form validation as well. I say, let's merge! Good job this was needed

PR #1116, opened 2024-08-02
Intro form overflow fix

SpookyBoy99:
The form on the intro page was not set up correctly with bootstrap causing it to lose its margins and be to wide which made the page overflow in the horizontal direction. Furthermore, on very small devices, the greeting header could be to large, making it unreadable and also causing the page to overflow.

### DIFF
--- a/app/assets/stylesheets/intro.css.scss
+++ b/app/assets/stylesheets/intro.css.scss
@@ -88,21 +88,32 @@ body {
 }
 
 .navbar {
-  .nav-item .nav-link {
-    color: #000000;
+  .container {
+    position: relative;
+  }
 
-    &.active {
-      color: #bb8400 !important;
-      font-weight: bold;
+  .navbar-nav {
+    @media (max-width: 576px) {
+      margin-top: 10px;
+      border-top: 1px solid #ced4da;
+    }
+
+    .nav-item .nav-link {
+      color: #000000;
+
+      &.active {
+        color: #bb8400 !important;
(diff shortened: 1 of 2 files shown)


SilasPeters (changes requested):
Although there are no overflow issues anymore (mission succesful), I do notice two things which are unfavourable on phone (you probably did not cause it, but I would like this to be fixed, and this PR is perfect for that).

image

In the image above, you can see that on most phones the Welcome text should be higher or lower. Furthermore, if it's not too much of a hassle, the localisation globe could be visible at (1), and not hidden in the menu - if an English person does not see the globe they quit quite quickly I believe.

If you can fix or argue against these things, then I am ready to approve (quick fix btw)!

SpookyBoy99:
I have fixed the language icon, I noticed this when I was developing the initial commit as well but I thought it would be non-trivial to fix. However, a little bit of `position: absolute;` magic on the mobile version and that is is now settled. 

I also do agree with the positioning of the text, but I'd rather leave it that way. There are too many phones with too many different aspect ratio's that it would be very hard to guarantee that different positioning would actually give the desired effect on the majority of the devices. Lastly, while I do think it is sad that the button is blocking the picture, the picture is there purely for aesthetic reasons and the call-to-action button is what should be the first thing that draws your attention (hence it is in your face in the middle).

SilasPeters (approved):
AEINMO! (Appearing Excellent To My Opinion)

PR #1118, opened 2024-08-14
Remove TRX from introduction payments

TobiasDeBruijn:


SilasPeters (approved):
I have not tested this, but the logic seems clear.

PR #1126, opened 2024-09-14
Removed code from the old Mongoose back-end and re-organised menu's

### LINKED ISSUE #1109: Remove Mongoose checkout code in the backend
Most if not all of the Mongoose code has been removed from the UI, but most is still present or commented out in the backend. This should be removed.

SpookyBoy99:
Closes #1109. The Mongoose data will not deleted from the database with this PR and Mongoose top-ups are still shown in the transaction history and exports.


SilasPeters (commented):
Seems fine at first glance, but we should carefully test all/most controllers of koala to see if nothing breaks

SpookyBoy99:
This PR makes no changes to the database so if something breaks we can easily revert to an earlier deployment and not data will be lost. I have looked through all the pages that I know of and they seem functional (including the export of transactions)

TobiasDeBruijn:
Looks good to me. @SilasPeters are there any blockers left on this PR?

SilasPeters:
I am not aware of any blockers. I would say let's review, test and merge! But @stickyPiston probably has a better view on this.

SpookyBoy99:
If @stickyPiston and @TobiasDeBruijn have time for a quick review, I can resolve the merge conflicts and merge this PR. I don't have a lot of time these days so if many more merge conflicts will arise in the mean time I don't think I will have enough time to resolve them in the future.

TobiasDeBruijn (approved):

app/controllers/admin/logs_controller.rb
```diff
@@ -0,0 +1,10 @@
+#:nodoc:
+class Admin::LogsController < ApplicationController
+  def index
+    @limit = params[:limit] ? params[:limit].to_i : 50
+
+    @pagination, @impressions = pagy(Impression.all.order(created_at: :desc),
+                                     items: params[:limit] ||= 50)
```
SilasPeters, line 7:
```suggestion
                                     items: @limit
```

SilasPeters (changes requested):
Quick scan. Our time would be better spend to test every endpoint in staging.

PR #1127, opened 2024-09-14
Added verbose (failure) logging to the seeding process

SpookyBoy99:
This PR also removes the seeding of Mongoose transactions, but those aren't necessary anymore anyway

PR #1128, opened 2024-09-14
Removed second empty KOALA_DOMAIN from sample.env

SpookyBoy99:
The empty `KOALA_DOMAIN` variable caused issues in the seeding of the payments, as 'ideal' transactions require this value to be set. This caused the database to be only be seeded with 'pin' transactions.

### DIFF
--- a/sample.env
+++ b/sample.env
@@ -75,16 +75,13 @@ MAILCHIMP_TEACHER_ID=
 # Used for when testing webhooks
 NGROK_HOST=http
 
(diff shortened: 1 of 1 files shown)


TobiasDeBruijn (approved):

SilasPeters:
It is said in #1127 the caused symptoms were solved, but not on my machine. It still says that most activities have an empty redirect URI, and halve of the seeding thus fails.

Issue #1125, opened 2024-09-10
wysiwyg Koala activities


SpookyBoy99:
Koala activity descriptions do support HTML, but it uses a simple plain textarea so it does not clearly signal the HTML support and may be hard to use for users without any web development knowledge. Koala already contains wysiwyg editors for the emails and board posts so this should not require a major overhaul to any of the front-end code.

Related issues: #1087 #1124

PR #1095, opened 2024-02-20
Add keybinds to admin members page

TobiasDeBruijn:
-  'e' (edit)
- 'ctrl+esc' (cancel edit)
- 'ctrl+enter' (save edit)
-  's' || '/' (search)

PR #1129, opened 2024-09-14
Added wysiwyg editor to the activity description fields

### LINKED ISSUE #1124: wysiwyg formatting for emails and board post does not equal to the output
### Expected behaviour

Line spacing, font size and font family and alike are similar to posted result or email.

### Actual behavior

Line spacing doubles and the font family and size are noticeably different in a published board post or email from the wysiwyg preview.

SpookyBoy99:
Implements #1125 and fixes #1124. It is not the most beautiful implementation but this was the most practical solution within the established ecosystem.

### DIFF
--- a/app/assets/stylesheets/admin/activities.css
+++ b/app/assets/stylesheets/admin/activities.css
@@ -149,3 +149,12 @@
 .activity_table-cell--pseudo-hidden {
     color: transparent;
 }
+
+.quill-editor .ql-editor {
+  min-height: 200px;
+}
+
+.ql-editor p, .ql-editor ol, .ql-editor ul {
+  margin-bottom: 1rem;
+  padding-left: 0;
+}
--- a/app/assets/stylesheets/members.css
+++ b/app/assets/stylesheets/members.css
@@ -11,6 +11,7 @@
  *= require yeti
  *= require font_awesome5_webfont
  *= require lato
+ *= require quill/dist/quill.snow
  *
  *= require intl_tel_input
  *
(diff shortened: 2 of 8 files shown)


SilasPeters (changes requested):
Great initiative. Even ctrl-z works! After testing, here are some remarks:

1. I can't enter the `s` character in the description field.
2. As you might have seen on whatsapp, the 'copy to whatsapp' does not convert/filter the HTML properly to markdown

SpookyBoy99:
Not being able to enter an `s` is probably a side effect from the PR #1095, it has some more issues unfortunately. The copy to WhatsApp filter I'll have to rewrite completely I'm afraid, probably using the Delta API from Quill. A quick solution for now however, would be to just copy the inner text which would just strip all the HTML.

SilasPeters:
@SpookyBoy99 what are your intentions on this PR?

SpookyBoy99:
It's not the greatest implementation of a WYSIWYG editor and the copy for WhatsApp functionality is pretty much broken. I am not sure if I have the time or will power to completely fix those issues, so I might just close this PR unless there is enough interest from the board.

TobiasDeBruijn:
While definitely a nice to have, the copy for whatsapp feature is a must have. If you feel like it, please do implement! Though if you'd rather focus on other projects or issues, that is completely fine as well, then this PR can be closed.

SpookyBoy99:
I'll see what I can do during the 14 hour bus ride this weekend, if I don't manage to get it working by then I will just close this PR

PR #1143, opened 2024-12-10
Added a function to age at date

CloevVeller:
Added the function that checks the user's age at the time of the event in stead of at the time of enrollment

### DIFF
--- a/app/models/member.rb
+++ b/app/models/member.rb
@@ -246,6 +246,11 @@ class Member < ApplicationRecord
     !adult?
   end
 
+  def underage_at?(event_date)
+    return birth_date.blank? ||
+       ((18.years.ago + (event_date - Date.today)) < birth_date)
+  end
+
   def master?
     educations.any? do |education|
       education.status == 'active' && Study.find(education.study_id).masters
--- a/app/controllers/members/participants_controller.rb
+++ b/app/controllers/members/participants_controller.rb
@@ -176,7 +176,7 @@ class Members::ParticipantsController < ApplicationController
   end
 
   def participant_alcohol_check?
-    @activity.is_alcoholic? && @member.underage?
+    @activity.is_alcoholic? && @member.underage_at?(@activity.start)
   end
 
   # [PATCH] /activities/:id/participants


SilasPeters (dismissed):
Seems to work at first sight! However, our pipeline complains. You should see the logs of the pipeline. Perhaps some checks failed or your code does not match our style requirements.

SilasPeters (commented):
I would approve the changes, but I want to test this beforehand (dates and timezones might give unexpected results). I'm quite busy though, so don't expect me to have done this in a day or two. Perhaps someone else could adjust (or expand?) the test suite to test this edge case (and, especially, if no one else unjustly can sign up).

app/models/member.rb
```diff
@@ -246,6 +246,12 @@ def underage?
     !adult?
   end
 
+  def underage_at?(event_date)
+    return true if birth_date.blank?
+
+    return ((18.years.ago + (event_date - Date.today)) < birth_date)
```
SilasPeters, line 252: I think `birth_date + 18.years < event_date` would read better (if that's still correct (code)).

SilasPeters:
I could not create a scenario where someone under the age of 18 could sign up for an activity with alcohol past their 18th birthday. Thus, I think this does not work.

Reproduction steps:

- Log in as `test@svsticky.nl` and note your name.
- Find an activity that is open to sign up for, note its id.
- Log out and back in as `dev@svsticky.nl`, and make your user underage.
- Set the activity's start and end date one day before the day your user would turn 18, and make the activity 18+.
- Assert that, when loggin in as test, you cannot sign up.
- Log back in as dev, and change the date to start a day later (on test's birthday).
- Logging back in as test, assert that you can sign up. <=== this fails. Same thing if the activity is in 2030.

SilasPeters (changes requested):
I could not create a scenario where someone under the age of 18 could sign up for an activity with alcohol past their 18th birthday. Thus, I think this does not work.

Reproduction steps:

- Log in as `test@svsticky.nl` and note your name.
- Find an activity that is open to sign up for, note its id.
- Log out and back in as `dev@svsticky.nl`, and make your user underage.
- Set the activity's start and end date one day before the day your user would turn 18, and make the activity 18+.
- Assert that, when loggin in as test, you cannot sign up.
- Log back in as dev, and change the date to start a day later (on test's birthday).
- Logging back in as test, assert that you can sign up. <=== this fails. Same thing if the activity is in 2030.

SilasPeters:
I now realise the error was 'Could not enroll' instead of 'Could not enrol, its for adults only'. Perhaps something weird is going on.

PR #1144, opened 2024-12-24
In dark mode, use the coloured logo as well

TobiasDeBruijn:
On the login page, in light mode we used the coloured logo, but in dark mode we used the white logo (due to the lack of a white *and* coloured logo). I've added this logo to public.svsticky.nl, and the login page now uses it.


SilasPeters (approved):
The filename is not in english smh

PR #1145, opened 2024-12-24
fix: finish ical feed

### LINKED ISSUE #1037: Generate and maintain user-specific calendars containing the corresponding users' activities

SilasPeters:
**I recommend reviewing per commit, as this PR includes the reversion of a reversion of a previous PR merge**

In #1101 I tried to close #1037, but on staging it did not seem to work. After some testing, this should have fixed it!

Aside from several improvements, I solved the issue where datetimes containing just dates - and no implicit information on how to deal with timezones - would be converted to utc. This was an issue, because if 10-10-2010 00:00 would be set a an hour back, it would become 09-10-2010 23:00, whilst we just wanted the date which now moved a day back.

Only thing to fix: importing in google calendar does not seem to work.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -49,6 +49,9 @@ gem 'pg'
 # Full text search
 gem 'pg_search'
 
+# iCalendar feeds
+gem 'icalendar'
+
 group :production, :staging do
   gem 'sentry-raven'
   gem 'uglifier'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -151,6 +151,9 @@ GEM
       rails-i18n
       rainbow (>= 2.2.2, < 4.0)
       terminal-table (>= 1.5.1)
+    icalendar (2.10.1)
+      ice_cube (~> 0.16)
+    ice_cube (0.16.4)
     image_processing (1.12.1)
       mini_magick (>= 4.9.5, < 5)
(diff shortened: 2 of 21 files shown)


SilasPeters:
Somehow, it works on my machine but not on staging... Full days start a day too early. Is this a timezone thing?

SilasPeters:
Note to self: misschien `VEVENT>CREATED` en `VEVENT>LAST-MODIFIED` gebruiken?

PR #1133, opened 2024-09-24
Update to Mollie v2 API

leuke-naam:
Closes #1130

PR #1147, opened 2024-12-26
Update to Mollie v2 API: Part 2 electric boogaloo

leuke-naam:
Fixes issues with #1133

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -27,6 +27,9 @@ gem 'impressionist', github: 'charlotte-ruby/impressionist'
 # rests calls for mailgun
 gem 'rest-client'
 
+# mollie
+gem 'mollie-api-ruby'
+
 gem 'responders'
 
 # pagination
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -179,6 +179,7 @@ GEM
     mini_mime (1.1.2)
     mini_portile2 (2.7.1)
     minitest (5.15.0)
+    mollie-api-ruby (4.13.0)
     netrc (0.11.0)
     nio4r (2.5.8)
     nokogiri (1.13.1)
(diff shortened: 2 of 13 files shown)


TobiasDeBruijn (dismissed):

SilasPeters (approved):
Lgtm

SilasPeters:
@leuke-naam and @TobiasDeBruijn I pulled a big brain and tested this branch on staging _without_ merging, by simply checking out this branch and restarting koala :brain: . Everything seems to work! Two things though:

1. If the payment is left as 'open', the user is not notified, but I believe @leuke-naam showed me once that a blue banner displaying 'your payment is being processed' should pop up. If that's true, that did not work.
2. The payments are in test mode, but I assume this is toggled off when a different `MOLLIE_TOKEN` is used in `.env` in production.

TobiasDeBruijn (approved):

leuke-naam:
> @leuke-naam and @TobiasDeBruijn I pulled a big brain and tested this branch on staging _without_ merging, by simply checking out this branch and restarting koala 🧠 . Everything seems to work! Two things though:
> 
> 1. If the payment is left as 'open', the user is not notified, but I believe @leuke-naam showed me once that a blue banner displaying 'your payment is being processed' should pop up. If that's true, that did not work.
> 2. The payments are in test mode, but I assume this is toggled off when a different `MOLLIE_TOKEN` is used in `.env` in production.

1. Payment status `open` is not applicable for iDeal payments, if in the future we were to accept more payment types we could handle this status
2. Correct

SilasPeters:
Then everything works! Somehow GitHub complains some requirements for merging are not met, so I will admin abuse

PR #1153, opened 2025-01-09
Changed English translation of VoelJeVeilig to 'Feel Safe'

KoenDR06:
As per [this](https://svsticky.nl/en-US/vereniging/vertrouwenspersonen) page, the English translation of 'Voel Je Veilig' is 'Feel Safe'.

In my opinion, this should be consistent.


SilasPeters (approved):
Highly agree

PR #1154, opened 2025-01-25
feat: update ruby and node versions

stickyPiston:
The original idea was to make koala mac-compatible, but it has turned into a general updating of a lot of packages and runtimes.


SilasPeters:
A bold, dangerous yet required move

SilasPeters:
Please make future changes in seperate commits so it's easier to review/understand what the purpose was of the changes

PR #1158, opened 2025-02-02
fix: migrate intro page to new mollie sdk

SilasPeters:
Basically removes the option to select a bank, which is deprecated.


SilasPeters:
Does not resolve #1159, but that's out of scope

SilasPeters (commented):
I tested this but again not fully because I don't use ngrok to make mollie work. @leuke-naam we really need documentation on this

stickyPiston:
Undead-mongoose has some documentation on how to use ngrok with it, so we can copy that and adapt it a little.

SilasPeters:
Should we perhaps merge now because of the first-year students activity tomorrow?

stickyPiston (approved):
頑張ったね！

stickyPiston:
It has been a massive pain to get this tested, adding account requires the mailer to work or have database access, neither of them particularly convenient. So we might want to make some utilities for that.

PR #1160, opened 2025-02-04
Add a payment deadline

TobiasDeBruijn:


app/views/members/activities/partials/_view.html.haml
```diff
@@ -68,6 +68,12 @@
         %span.activity-unenroll
           = l(activity.unenroll_date, format: "%A %d/%m/%Y")
           23:59
+        <br/>
+      - if !activity.payment_deadline.nil?
+        %strong= I18n.t('activerecord.attributes.activity.payment_deadline')
+        %span.activity-payment-deadline
+          = l(activity.payment_deadline, format: "%A %d/%m/%Y")
+          23:59
```
stickyPiston, line 76: Do we want to display the payment deadline to the users? Or maybe only display it when it differs from the default?

db/migrate/20250204135006_add_payment_deadline_to_activities.rb
stickyPiston, line 1: Could you change the seeding to also generate some activities with payment deadlines?

app/models/activity.rb
```diff
@@ -113,6 +118,8 @@ def self.debtors
     joins(:participants).where('
       activities.is_payable
       AND
+      (activities.payment_deadline IS NULL OR activities.payment_deadline > ?)
```
stickyPiston, line 121: Shouldn't it be `activities.payment_deadline <= ?`? When today is later than the payment deadline, you're a debtor I would think. Now, if I set the payment deadline for an activity to somewhere in the future, it will still appear on the admin payments page, whereas setting it in the past will make it disappear.

stickyPiston (changes requested):

app/views/members/activities/partials/_view.html.haml
TobiasDeBruijn: Yeah why not. The default is pay whenever the treasurer gets angry, but always after the activity. For the studytrip and airco this differs. I think only showing it when it is set (and thus when it is relevant) is the best option.

app/models/activity.rb
TobiasDeBruijn: that is a good point. That mistake is elsewhere too.

app/views/members/activities/partials/_view.html.haml
stickyPiston: Alright, makes sense

stickyPiston (approved):
頑張ったね！

PR #1165, opened 2025-02-28
fix url routing intro

SamStaijen:


stickyPiston (approved):
頑張ったね！

SilasPeters (commented):
How about `join.svsticky.nl` for the non-dutch members?

stickyPiston:
Ohhh good one, we could discuss this with @olafboekholt next week. It shouldn't be difficult to add another alias for that one.

olafboekholt:
> Ohhh good one, we could discuss this with @olafboekholt next week. It shouldn't be difficult to add another alias for that one.

Currently we already have `signup.svsticky.nl`, but adding `join.svsticky.nl` actually makes a lot of sense!

SilasPeters:
See the latest issue :)

PR #1169, opened 2025-03-31
feat: display member name in nav bar

stickyPiston:


SilasPeters (commented):
Seems like it must be correct. I'm uncertain if credentials.X might sometimes contain secret info and not just a name, have you sufficiently tested this before we expose everyone's credentials?

SilasPeters (approved):
Works on my machine.

SilasPeters:
We could merge this @stickyPiston

PR #1170, opened 2025-04-01
Clippy eruit geklappie

SamStaijen:
revert van laatste commit, want CTF is geweest


stickyPiston:
The commits have already been reverted and force-pushed so that the history isn't contaminated with the clippy-related commits.

PR #1171, opened 2025-04-04
fix: include other signup domains

SilasPeters:
Koala was not aware of our new domains, and as such does not accept authentication tokens from those domains.

Starting a local version of those domains should now also be possible.


SamStaijen (approved):
LGTM

PR #1173, opened 2025-04-25
fix: new member activation link pointing to confirmation url

SilasPeters:
A little is better than nothing. Me and Tobias have now tried to solve the bug's cause, in vein. Thus, here is a workaround.

The intention is that a noticable explenation is present in the email, explaining what the user should do if they have the incorrect link.

I have not tested what this looks like, for I don't know how to test emails in staging environments (non are send).

@TobiasDeBruijn do you agree?
@olafboekholt are the included instructions correct?

### DIFF
--- a/config/locales/mailings.en.yml
+++ b/config/locales/mailings.en.yml
@@ -4,15 +4,15 @@ en:
     best_regards: Best regards,
     checkout:
       confirm_card: Confirm card
-      confirm_card_link: Confirme Checkout card for your account at Study association Sticky by going to %{confirm_url}.
+      confirm_card_link: Confirm Checkout card for your account at Study association Sticky by going to %{confirm_url}.
       confirmation_instructions: 'Confirm your card for your Checkout account with Study association Sticky by clicking the following button:'
       subject: Confirm Checkout card
     confirm: Are you sure you want to sent this email?
     devise:
       activation_instructions:
-        about_sticky: 'You can go to Sticky for everything that you as Computer Science, Information Science or Game technology need! As you might have noticed already, you can always score free coffee and a cookie in our room (BBG 2,81) or some other other drink/snack. There will always be someone of the board present in the room who can answer all your questions. Besides this place to chill during breaks or in between lectures, we also organise a whole lot of wildly awesome activities. These activities These activities are always aligned to one of the three corner stones of Sticky: Education, Business and Sociability.'
-        account_activation_instructions: Activate your account by clicking the button below
-        account_activation_link: Activate your account for our member system by going to %{url}.
+        about_sticky: 'You can go to Sticky for everything that you as Computer Science, Information Science or Game technology student need! As you might have noticed already, you can always score free coffee and a cookie in our room (BBG 2,81) or some other other drink/snack. There will always be someone of the board present in the room who can answer all your questions. Besides this place to chill during breaks or in between lectures, we also organise a whole lot of wildly awesome activities. These activities These activities are always aligned to one of the three corner stones of Sticky: Education, Business and Sociability.'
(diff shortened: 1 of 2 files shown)


config/locales/mailings.en.yml
```diff
@@ -4,15 +4,15 @@ en:
     best_regards: Best regards,
     checkout:
       confirm_card: Confirm card
-      confirm_card_link: Confirme Checkout card for your account at Study association Sticky by going to %{confirm_url}.
+      confirm_card_link: Confirm Checkout card for your account at Study association Sticky by going to %{confirm_url}.
       confirmation_instructions: 'Confirm your card for your Checkout account with Study association Sticky by clicking the following button:'
       subject: Confirm Checkout card
```
TobiasDeBruijn, line 9: Sidenote: This can be removed entirely considering it is no longer used.

config/locales/mailings.en.yml
```diff
@@ -4,15 +4,15 @@ en:
...
-        account_activation_instructions: Activate your account by clicking the button below
-        account_activation_link: Activate your account for our member system by going to %{url}.
+        about_sticky: 'You can go to Sticky for everything that you as Computer Science, Information Science or Game technology student need! As you might have noticed already, you can always score free coffee and a cookie in our room (BBG 2,81) or some other other drink/snack. There will always be someone of the board present in the room who can answer all your questions. Besides this place to chill during breaks or in between lectures, we also organise a whole lot of wildly awesome activities. These activities These activities are always aligned to one of the three corner stones of Sticky: Education, Business and Sociability.'
+        account_activation_instructions: Activate your account by clicking the button below.
+        account_activation_link: Activate your account for our member system by going to %{url}. NOTE if this url does not work, replace 'confirmation' at the end of the url with 'activate'! Why this can go wrong, you ask? Long story.
```
TobiasDeBruijn, line 15: I think an example could explain it further; so an example URL

config/locales/mailings.nl.yml
```diff
@@ -12,7 +12,7 @@ nl:
       activation_instructions:
         about_sticky: 'Bij Sticky kun je terecht voor alles wat jij als student Informatica, Informatiekunde of Gametechnologie nodig hebt! Zoals je wellicht al vernomen hebt, kun je altijd een gratis kop koffie en een koekje scoren in onze kamer (BBG 2.81), of een ander(e) drankje/snack. In de kamer is ook altijd iemand van het bestuur aanwezig om al jouw vragen te beantwoorden. Naast deze plek om te chillen tijdens pauzes of tussen de colleges door, organiseren we ook nog eens enorm veel woestgave activiteiten! Deze activiteiten zijn altijd in lijn met één van de drie pijlers van Sticky: onderwijs, bedrijfsleven en gezelligheid.'
         account_activation_instructions: 'Activeer je account door op onderstaande knop te klikken:'
-        account_activation_link: Activeer je account voor ons ledenbeheersysteem door naar %{url} te gaan.
+        account_activation_link: Activeer je account voor ons ledenbeheersysteem door naar %{url} te gaan. LET OP! Als deze link niet werkt, vervang dan 'confirmation' aan het einde van de link met 'activate'! Waarom dit soms nodig is? Lang verhaal.
```
TobiasDeBruijn, line 15: Same story regarding the example, but then in Dutch

TobiasDeBruijn (changes requested):

TobiasDeBruijn (approved):

TobiasDeBruijn:
hopefully with the chatter in the whatsapp channels, this fix won't even be necessary

SilasPeters:
#1174 is a better solution, it just straight up removes the activate link and makes everything work.

PR #1174, opened 2025-04-30
feat: remove old registration form

stickyPiston:
The old registration form causes a lot of confusion among people who want to sign up for sticky. Since we only want to allow signing up from wordlid.svsticky.nl, I reckon it is wise to remove the old registration form. The additional bonus is that we remove one source of sending emails with the wrong activation link.

### DIFF
--- a/app/controllers/users/registrations_controller.rb
+++ b/app/controllers/users/registrations_controller.rb
@@ -5,46 +5,6 @@ class Users::RegistrationsController < ActionController::Base
 
   layout 'doorkeeper'
 
-  def new
-    @user = User.new
-    render('devise/registrations/new')
-  end
-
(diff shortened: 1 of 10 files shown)


TobiasDeBruijn (dismissed):

SilasPeters (commented):
This PR is not a draft, but you mentioned that you would rewrite logic such that one endpoint would function for both changing email and confirming a new email, but I do not see that logic change yet.

stickyPiston:
No you're right, I'm still working on that! I'll set this PR back to a draft PR.

SilasPeters:
What's the progress so far? Can we help? This is quite crucial for new members arriving every week at this point

PR #1175, opened 2025-05-06
fix: Hyphens in names of members are deleted

### LINKED ISSUE #1161: Hyphens in names of members are deleted
Some people have a hyphen (-) in their name, but when signing up or changing the name in the members admin view, the hyphen is removed. The issue lies in the call to `titleize` in the following code. https://github.com/svsticky/constipated-koala/blob/718c024338d9c0cf169f988420034854ad61a7f9/app/models/member.rb#L102-L105

KoenDR06:
Closes #1161


stickyPiston (approved):
頑張ったね！

PR #1176, opened 2025-05-06
Copy to whatsapp price rounding

### LINKED ISSUE #1167: Copy to whatsapp price rounding
Prices in the copy to whatsapp are rounded to one decmical, but since it is money, it should be two.

See the picture, would be nice if it was 10.00 or just 10 without decimal. If an activity is 9.50, it now shows 9.5 which is weird when it comes to currency.
Image

KoenDR06:
Closes #1167


stickyPiston (approved):
頑張ったね！

PR #1186, opened 2025-09-07
Fix activities list dropdown (#833)

### LINKED ISSUE #833: Selecting another year of activities blocks returning to the current year
### Expected behaviour
When selecting another year in the activites box on the koala homepage (after logging in) should load the correct year page, but also display the selected year in the dropdown menu.

### Actual behavior
The selected year is overwritten with the current year, but clicking that year does not return the user to the current year page.

### Steps to reproduce
- user
- browser: Firefox 89 (but probably also on all other browsers)

When on the home page, select another year in the year selector (with class "year-selector"). The page then loads the activities of the user in that year, but the selector resets to the current year.

NoahvdAa:
Closes #833.
Every other year picker reads the selected year except this one, so this was a very easy fix :)


SilasPeters (approved):
Simple yet effective

SilasPeters:
Next time we deploy the latest changes to production, this will be included!

PR #1216, opened 2026-04-28
Fix: include start_time in activity sort order

### LINKED ISSUE #1215: Board activity list wrong sort order on same-day activities
In the weekoverzicht, activities on the same day should be sorted ascending. Now they are sorted descending (or maybe another sort, didn't check this).

Edit: turns out this is also the case in the global activity list on the board view. Weekoverzicht probably copies this.

Quintus3:
Closes #1215.
Included start_time in the sorting arguments, instead of only start_date.

### DIFF
--- a/app/controllers/admin/activities_controller.rb
+++ b/app/controllers/admin/activities_controller.rb
@@ -65,7 +65,7 @@ https://koala.svsticky.nl/activities/#{ ac.id }
       impressionist(@activity)
       redirect_to(@activity)
     else
-      @activities = Activity.all.order(start_date: :desc)
+      @activities = Activity.all.order(start_date: :desc, start_time: :desc)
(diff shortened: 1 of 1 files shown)


SilasPeters:
Lol what do you think of #1217 ? He has one extra place where this fix takes place, not sure where one is needed.

Quintus3:
> Lol what do you think of #1217 ? He has one extra place where this fix takes place, not sure where one is needed.
@SilasPeters 

I've looked at his change, and it seems like lines 68-73 are just copies from lines 6-9, which is being called afterwards anyways. Correct me if I'm wrong, but I think lines 68-73 can just be replaced by `index()`.
If this is incorrect, the proposed changes by that other guy seem fine, thus adding `, start_time: desc` to line 68.

SilasPeters (dismissed):

SilasPeters (approved):

